PR #2, opened 2024-03-26
basic package struct

Teschl:


wkearn:
Looks great, Theo!

I think we probably don't want to commit the build artifacts in `dist/`. We'll eventually release TopoToolbox in PyPI, and we can add the binaries to the [releases](https://github.com/TopoToolbox/pytopotoolbox/releases/) page here when we start making releases.

PR #11, opened 2024-04-10
Basic GridObject and .info() function

Teschl:
I also removed the `examples/` from `.gitignore`


wkearn:
This looks good. We'll probably want to address @wschwanghart's comments on #6 at some point, but if we need to make changes based on those comments, it shouldn't be too hard at this point to add things to the current version of `GridObject`.

The one thing we probably should figure out now is the name: `GridObject` as we have it here or `GRIDobj` like the MATLAB version or something else? I assume that the argument for `GridObject` is that it sounds more like other Python classes, right? Why not drop the `Object` and call it `Grid`?

Teschl:
Anna-Lena always preferred the `GridObject` variant to `GRIDobj` since it's the "pythonic" variant as specified in [PEP 8](https://peps.python.org/pep-0008/#class-names).

We could also add aliases in the `__init__.py` file. That way, `GridObject`, `Grid` or 'GRIDobj` would all work, and the user can pick whichever they like best.

PR #14, opened 2024-04-20
added fillsinks with pybind11

Teschl:
included flow.cpp as showcase of future classes


src/topotoolbox/gridmixins/fillsinks.py
```diff
@@ -0,0 +1,18 @@
...
+    def fillsinks(self):
+
+        dem = self.z.astype(np.float32)
+
+        output = np.zeros_like(dem)
+
+        grid_fillsinks(output, dem, self.rows, self.columns)
+
+        result = copy.copy(self)
+        result.z = output.copy()
```
wkearn, line 16: I might be wrong, but I don't think you need this copy. If you tried to assign `result.z[0,0]  = output[0,0]` and so on, you would overwrite the elements of `self.z`, which we don't want to do. However, I believe `result.z = output` just makes `result.z` point at `output` without touching `self.z`, so it should be safe without the copy.

wkearn (changes requested):
Looks great, @Teschl!

I added a couple comments. The one on `grid.cpp` is more for future reference, so don't worry about changing anything there. Take a look at the one on `fillsinks.py` about a possibly unnecessary copy and either let me know if there is something I am missing or remove that copy. Then I think we should be able to merge this.

src/topotoolbox/gridmixins/fillsinks.py
Teschl: I think you may be right. The use `.copy()` seems to be an artifact from earlier implementations. We should be able to safely remove it.

wkearn (approved):

PR #15, opened 2024-04-20
Added GridObj generation with OpenSimplex noise

Teschl:
Useful to better visualize changes when using `fillsinks()` for example. I used [opensimplex](https://pypi.org/project/opensimplex/) to compute the array. I don't think it's necessary to implement Simplex or Perlin noise in `libtopotoobox` since it's only use is for testing or examples. Nevertheless, I think it's a nice addition to have. `opensimplex` will not be automatically installed, but the user will be prompted to install it when trying to use `topotoolbox.GridObject.gen_random()`.

We'll probably also need a better name than `gen_random` since it's not really random.


wkearn:
Neat! I was thinking about adding some kind of gradient noise generator to the testing apparatus for `libtopotoolbox`, but I think you are right that it doesn't really make sense to implement it in the library itself at least for now.

Does it make sense to declare `opensimplex` an [optional dependency](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-optional-dependencies) of the package? This could be useful going forward if we want to have users install a few packages just to run the examples.

Teschl:
I added the optional dependency to the `pyproject.toml` and modified the error message when trying to use `gen_random()` to prompt the user to install with `opensimplex`. Right now its `pip install .[opensimplex]`. I think we can go ahead and merge if you want @wkearn.

PR #21, opened 2024-05-10
Add pybind11 license to LICENSES_bundled

### LINKED ISSUE #20: Handle PyBind11 License
Since PyBind11 is using a BSD-style license that, among other things, states: 

"2. Redistributions in binary form must reproduce the above copyright notice,
   this list of conditions and the following disclaimer in the documentation
   and/or other materials provided with the distribution."
[from pybind11 license](https://github.com/pybind/pybind11/blob/master/LICENSE)
   
Does that mean we also have to include the license in our repository? If so, we'll probably need to check all other dependencies for similar requirements.

wkearn:
Fixes #20 

This should comply with the 3-clause BSD license used by pybind11.

@Teschl Can you confirm that if you run `python3 -m build` that `LICENSES_bundled` shows up in the source code archive and wheel that get built?

### DIFF
--- a/LICENSES_bundled
+++ b/LICENSES_bundled
@@ -1,0 +1,35 @@
+pybind11
+
+While no source code for pybind11 is included in pytopotoolbox,
+pybind11 is included in the shared libraries provided by the binary
+builds of pytopotoolbox.
+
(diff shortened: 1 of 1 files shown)


Teschl:
I can confirm, LICENSES_bundled does show up in the build archive.

PR #23, opened 2024-05-17
Implement Pep8 standard

Teschl:
Added Docstrings to all python files, enforced max line lengths, and fixed issued pointed out by pylint. Also added comments to `grid.cpp`.


wkearn:
This looks good. Thanks, @Teschl!

PR #24, opened 2024-05-17
Add Identifyflats

### LINKED ISSUE #22: Add identifyflats
The basic implementation should be not much different from `fillsinks`, but the output array is made up of `int32_t `and not `float`.
But instead of just one, there needs to be two (three?) instances of `GridObject `returned.
- `flats`
- `sills`
- `presills` (is that the same as CLOSED in matlab? @wkearn )

These have to be generated from the output matrix.

Depending on if we want to add an argument that decides which of these will get returned, we can decide between something that look _roughly_ like one of those:
1. `flats, sills, presills = dem.identifyflats()`
alternatively `result = dem.idenifyflats()` where `result = (flats, sills, presills)`

2. `flats= dem.identifyflats(output="flats")`, 
`flats, sills = dem.identifyflats(output=["flats", "sills"])` and so on. 

Maybe both at the same time, with 1. being the standard would work too.

Teschl:
Implements `identifyflats` for the `GridObject`. To be able to use this function efficiently when creating the `FlowObject` I added an argument `raw` to just return the `ndarray`.

Since there is no way of making a function private in python, I think it's important to still add a function that returns multiple `GridObjects`. By using the `output` argument, the number of returned dems can be changed.

- `flats, sills = dem.identifyflats(output=['flats', 'sills'])` (default)
- `flats = dem.identifyflats(output=['flats'])`
- `sills = dem.identifyflats(output=['sills'])`

image

When using `dem.identifyflats(raw=True)` instead of multiple `GridObject` the return will be an int32_t `nd.ndarray`.  This also contains presills.

image

Closes #22


src/topotoolbox/gridmixins/identifyflats.py
```diff
@@ -0,0 +1,49 @@
...
+            raw (bool): If True, returns the raw output grid as np.ndarray. 
+                Defaults to False.
+            output (list): List of strings indicating desired output types.
+                Possible values are 'sills', 'flats'. Defaults to ['sills', 'flats'].
+
+        Returns:
+            tuple: A tuple containing copies of the DEM with identified flats and/or sills.
+
+        Note:
+            Flats are identified as 1s, sills as 2s and presills as 3s in the output grid.
```
wkearn, line 24: Presills are labeled as 5 in the output grid ( 5 = 0b001 | 0b100 ) so they show up as both flats and presills.

wkearn (changes requested):
This looks pretty good! I think having different return types for external and internal use is a good idea for now.

There is one problem with the documentation for `identifyflats`.

You also need to merge/rebase the latest changes to `main`: the comments introduced by #23 to `grid.cpp` are causing a merge conflict.

Teschl:
Adjusted the documentation and resolved the merge conflict, should be ready to be merged now. @wkearn

wkearn (approved):

PR #25, opened 2024-05-22
Windows Installation Guide

Teschl:
Reorganized Readme.md structure. The Linux Guide is mostly the same as before.

1. **For Windows:**
It took quite a while to get Windows to build the package while not crashing during the C/C++ compiling stage. But now that I have got it running, it's pretty straight forward. To make sure that the guide actually works, it would be nice if one of you could follow the steps in the guide, build the package and then just run one of the example files to verify functionality. 

2. **For Mac**
Since I don't own an Apple Laptop, I can't check if the installation works there too. I assume it works the same as on Linux, but I have no way of knowing. I'll add an issue to track this.  (#26 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,65 +8,66 @@ Python interface to TopoToolbox.
 
 - **Installing from .whl file**
 
-    Make sure to choose the appropriate file for your OS. For Linux the file name should contain something like: `linux_x86_64`
+	Make sure to choose the appropriate file for your OS. For Linux, the file name should contain something like: `linux_x86_64`
 
-    ```bash
-    pip install dist_name.whl
-    ```
+	```bash
+	pip install dist_name.whl
+	```
 
-- **Installing directly from the reposetory:**
+- **Installing directly from the repository:**
 
-    ```bash
-    cd path/to/pytopotoolbox
-    pip install .
-    ```
+	```bash
+	cd path/to/pytopotoolbox
+	pip install .
+	```
 
 - **Generating distribution archives**
 
-    ```bash
-    cd path/to/pytopotoolbox
-    python3 -m pip install --upgrade build
-    python3 -m build
-    ```
+	```bash
+	cd path/to/pytopotoolbox
+	python3 -m pip install --upgrade build
+	python3 -m build
+	```
 
 ### Windows
 
 - **Installing from .whl file**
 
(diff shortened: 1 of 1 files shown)


wkearn:
Once I get back into the office, I will try to run through this on my Windows computer. I will also try to set up a Windows build on GitHub Actions (#27) and see how that goes.

The Windows build instructions are unfortunately complicated, but I guess most users of pytopotoolbox will use `pip install topotoolbox` or something and get a precompiled binary downloaded from PyPI, right? Once we have published it to PyPI of course. 

It's only developers of the package or people who really want to build the code locally who would need to go through this process, and if they are developing on Windows they probably already have Visual Studio installed.

Actually, that does bring up a question: what happens on Linux if you don't have CMake installed? Does scikit-build-core bring its own CMake? If not, it might be useful to mention that dependency.

Teschl:
> The Windows build instructions are unfortunately complicated, but I guess most users of pytopotoolbox will use pip install topotoolbox or something and get a precompiled binary downloaded from PyPI, right? Once we have published it to PyPI of course.

When we upload to PyPI it will be as a `.whl` file, so the whole Visual Studio stuff is only necessary when compiling the repository and building the `.whl`. 

> Actually, that does bring up a question: what happens on Linux if you don't have CMake installed? Does scikit-build-core bring its own CMake? If not, it might be useful to mention that dependency.

As far as I can tell, when cmake is not installed before running the build process, scikit-build-core will automatically install everything that is needed for building.

wkearn:
The Windows instructions did work for me. 

You do need to have `git` available on your path because CMake's FetchContent module attempts to clone our libtopotoolbox repository from GitHub. This is true for all operating systems, and it might be useful to mention that somewhere in these instructions.

At some point we could consider changing the pytopotoolbox build step so it doesn't depend on Git. I think that will make more sense once we have a consistent release process for libtopotoolbox so that we could download the appropriate release from somewhere.

Teschl:
I moved the general install requirements above the OS specific ones. 
> At some point we could consider changing the pytopotoolbox build step so it doesn't depend on Git. I think that will make more sense once we have a consistent release process for libtopotoolbox so that we could download the appropriate release from somewhere.

As far as I understand, CMake can download tar.gz files without using git, so if we provide one with each release this problem should be easy to solve.

Issue #9, opened 2024-04-04
Provide example data for testing and documentation


wkearn:
The example datasets are valuable to have around for testing and documentation, but we should generally avoid committing large binary files to the repository if we can. I think we need a system where we can store example data files outside the repository and download them either when the package is installed or when the user tries to access them.

Teschl:
I was thinking we could handle example datasets like seaborn. [load_dataset()](https://seaborn.pydata.org/generated/seaborn.load_dataset.html) will load an example dataset from the online repository.

> This function provides quick access to a small number of example datasets that are useful for documenting seaborn or generating reproducible examples for bug reports. It is not necessary for normal usage.

Using the [example DEMs](https://github.com/wschwanghart/DEMs) we could add functions equivalent to: `seaborn.load_dataset()` and `seaborn.get_dataset_names()`. We still should keep the random generation, so one can observe changes on a smaller, more visible scale. 
I'll work on getting a pull request ready for this.

PR #29, opened 2024-05-28
Provide example data

Teschl:
In this pull request, I added the following functions:

- `topotoolbox.load_dem()` enter DEM name as argument, it will download the tif if it isn't already cached.
- `topotoolbox.get_dem_names()` will return a list of all available tifs for `load_dem()`. if we add a .txt which contains all names to the repository which contains the tifs, we can implement it more dynamically. (right now it's just a list in the .py file)
- `topotoolbox.read_tif()` is a wrapper for creating a GridObject. Could be a cleaner way to create a GridObject inspired by pandas (pandas.read_csv creates a pandas.Dataframe).

Since the function `get_save_location()` in `utils.py` is not listed in `__all__` it's not automatically imported when using the package. This way it's "private" and not usable for the user by default.

I also fixed some line length issues in the mixin files for the GridObject. The maximum line length is 79 chars.

### DIFF
--- a/src/topotoolbox/utils.py
+++ b/src/topotoolbox/utils.py
@@ -3,15 +3,14 @@
 import sys
 import os
 
-from urllib.request import urlretrieve
+from urllib.request import urlopen, urlretrieve
 
 from .grid_object import GridObject
 
 __all__ = ["load_dem", "get_dem_names", "read_tif"]
 
-DEM_SOURCE = "https://raw.githubusercontent.com/wschwanghart/DEMs/master"
-DEM_NAMES = ['kunashiri', 'perfectworld',
-             'taalvolcano', 'taiwan', 'tibet', 'kedarnath']
+DEM_SOURCE = "https://raw.githubusercontent.com/TopoToolbox/DEMs/master"
+DEM_NAMES = f"{DEM_SOURCE}/dem_names.txt"
 
 
 def read_tif(path: str) -> GridObject:
@@ -28,17 +27,15 @@ def read_tif(path: str) -> GridObject:
 
 def get_dem_names() -> list[str]:
     """Returns a list of provided example Digital Elevation Models (DEMs).
+    Requires internet connection to download available names.
 
     Returns:
-        list: A list of strings, where each string is the name of a DEM.
-
-    Note:
-        If a file with all names is added to the 'wschwanghart/DEMs' 
-        repository, the DEM_NAMES list could be generated dynamically from 
-        that file. This would ensure the list remains up-to-date if the 
-        files ever change.
+        list[str]: A list of strings, where each string is the name of a DEM.
     """
-    return DEM_NAMES
+    with urlopen(DEM_NAMES) as dem_names:
+        dem_names = dem_names.read().decode()
+
+    return dem_names.splitlines()
 
 
 def load_dem(dem: str, cache=True) -> GridObject:
@@ -54,6 +51,10 @@ def load_dem(dem: str, cache=True) -> GridObject:
     Returns:
         GridObject: A GridObject generated from the downloaded dem.
     """
+    if dem not in get_dem_names():
+        err = ("DEM has to be chosen from the list of available DEMs."
+               "Use 'get_dem_names()' to generate list of all available DEMs")
+        raise ValueError(err)
 
     url = f"{DEM_SOURCE}/{dem}.tif"
 
--- a/src/topotoolbox/utils.py
+++ b/src/topotoolbox/utils.py
@@ -1,0 +1,109 @@
+import sys
+import os
+
+from urllib.request import urlretrieve
+
+from .grid_object import GridObject
+
+__all__ = ["load_dem", "get_dem_names"]
+
+DEM_SOURCE = "https://raw.githubusercontent.com/wschwanghart/DEMs/master"
+DEM_NAMES = ['kunashiri', 'perfectworld',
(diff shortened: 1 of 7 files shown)


Teschl:
Regarding the repository that contains the examples, right now, it's the [https://github.com/wschwanghart/DEMs](https://github.com/wschwanghart/DEMs) repository. So we might want to fork it to `https://github.com/TopoToolbox/DEMs`. Then we also should add a file like `contents.txt` in which we store all names of the provided DEMs. That means we could remove the hardcoded `DEM_NAMES` of `utils.py` and generate them dynamically.

If implemented, this should resolve #9 .

wkearn:
https://github.com/TopoToolbox/DEMs

src/topotoolbox/utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+    return dem
+
+
+def get_save_location() -> str:
+    """Generates filepath to file saved in cache.
+
+    Args:
+        data_home (str, optional): name of directory in cache. Defaults to None
+        which results in "topotoolbox" as name.
```
wkearn, line 76: This argument doesn't exist.

src/topotoolbox/utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+    url = f"{DEM_SOURCE}/{dem}.tif"
+
+    if cache:
+        cache_path = os.path.join(get_save_location(), f"{dem}.tif")
+
+        if not os.path.exists(cache_path):
+            urlretrieve(url, cache_path)
+
+        full_path = cache_path
```
wkearn, line 62: Can we provide a `clear_cache` function or something similar to allow users to empty this directory if they decide they don't need the data anymore?

src/topotoolbox/utils.py
```diff
@@ -0,0 +1,98 @@
...
+def read_tif(path: str) -> GridObject:
+    """Generate a new GridObject from a .tif file.
+
+    Args:
+        path (str): path to .tif file.
+
+    Returns:
+        GridObject: A new GridObject of the .tif file.
+    """
+    return GridObject(path)
```
wkearn, line 25: Would it make sense to read the data through rasterio in here and then use `read_tif` in the `GridObject` constructor? That might let us use the `GridObject` constructor to dispatch to different methods for loading data depending on the arguments provided to it. That might be a problem for another pull request though.

wkearn (changes requested):
Looks pretty good, @Teschl. This should be sufficient for our testing and documentation purposes at the moment. I suggested a few changes that make sense right now. 

In the future, we might want to think about other ways that users might want to access data and how we can accomodate them. The MATLAB TopoToolbox [provides](https://raw.githubusercontent.com/TopoToolbox/topotoolbox/2ef445126fd0e6c9c336dce19a205b2a2e62f8a2/IOtools/readopentopo.m), for example, a way to access the [OpenTopography API](https://opentopography.org/developers) that could be nice to have across our packages.

wschwanghart:
Excellent. I think it's generally a good idea to host also the example DEMs
on git although git is not really made for data hosting. However, so far it
supports file sizes up to 50 Mb. We might also consider this here:
https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-git-large-file-storage

Am Di., 28. Mai 2024 um 19:57 Uhr schrieb William Kearney <
***@***.***>:

src/topotoolbox/utils.py
Teschl: Yes, I think that would be better. I think moving the `load_dem` to `grid_object.py` would be a good move. Also, moving `gen_random()`, [...] out of the class will improve usability. (calling `topotoolbox.gen_random()` is better than calling `topotoolbox.GridObject.gen_random())

Therefore, calling the `__init__()` of the GridObject should require passing all attributes and having them prepared beforehand.

These are quite a few changes, though, so that should definitely be done in a different pull request.

Also, just to add to that: while implementing this pull request, I was looking at a few other repositories like `pandas` and `seaborn`. Nobody seems to be using mixins. It seems to be convention to have very long files (2000 lines is not uncommon like [here](https://github.com/pandas-dev/pandas/blob/main/pandas/core/arrays/categorical.py)). So after all, maybe we should get rid of them.
The only issue I see with that in our case is, that we are importing some files that only exist when building (`from ..grid import grid_identifyflats`). So my IDE is always yelling at me that some files don't exist. That's circumnavigated by using mixins.

src/topotoolbox/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    Args:
+        dem (str): Name of dem about to be downloaded
+        cache (bool, optional): If true the dem will be cached. 
+        Defaults to True.
+
+    Returns:
+        GridObject: A GridObject generated from the downloaded dem.
+    """
+
+    url = f"{DEM_SOURCE}/{dem}.tif"
```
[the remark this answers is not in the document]
Teschl: I was thinking we could also add a function like `store_dem()` that caches DEMs as .tif files. That will require the opposite of `read_tif()`: `write_tif()`.
Then calling `load_dem()` will either get the file with the respective name from the cache or download it if it's available and not in the cache already.

Teschl:
I have addressed the more bug-fix related requested changes. 

> Would it make sense to read the data through rasterio in here and then use read_tif in the GridObject constructor? That might let us use the GridObject constructor to dispatch to different methods for loading data depending on the arguments provided to it. That might be a problem for another pull request though.

and maybe adding `store_dem()` can be handled in a different pull request.

PR #39, opened 2024-06-17
Refactor package 

### LINKED ISSUE #34: Refactoring of package structure
Since creating the Sphinx documentation, it has become clear that using mixins is not common when writing python packages and makes life much harder when trying to document the code. It's common to files that contain 3000 lines of code, so having all functions in just a few files should be the goal of this. 
Generating a new `GridObject` also has to be streamlined to not require `topo.GridObject.gen_random()` instead of just `topo.gen_random()` for example.

Teschl:
I refactored the package, so the locations of different functions make more sense. The functions creating a `GridObject` are now in `utils.py`, and calling `topotoolbox.GridObject()` will result in an empty `GridObject`.

I also changed the way autosummary is used, so the way the functions are listed reflects the way they are used. Instead of having `topotoolbox.utils.load_dem()` in the documentation, it's now `topotoolbox.load_dem()` for example. The docstrings are now in the numpy format instead of the Google one, this keeps readability while insuring they are formatted right in the documentation.

closes #34


src/topotoolbox/grid_object.py
```diff
@@ -1,147 +1,341 @@
...
-class GridObject(
-        InfoMixin,
-        FillsinksMixin,
-        IdentifyflatsMixin,
-        MagicMixin
-):
+class GridObject():
     """A class containing all information of a Digital Elevation Model (DEM).
     This class combines mixins to provide various functionalities 
     for working with DEMs.
```
wkearn, line 18: This line is no longer true.

wkearn (changes requested):
Good work, @Teschl. I left one comment about some outdated documentation.

When I went to build the docs, I got some errors because my old `_autosummary` directory still had files from the older version. I guess `make clean` just cleans the `_build` directory. Do you know of a way to make sure that those `_autosummary` files also get deleted automatically? It's not a big deal, because I can always manually delete that directory. It's just a little annoying to have to remember to do that.

Teschl:
@wkearn I added 'clean' to the makefile, overwriting the Sphinx version. It now cleans `_build`, `_examples` and `_autosummary`. After building, the `_autosummary` directory will get cleaned as well.

PR #51, opened 2024-07-18
Fix matrix passing to C/C++

### LINKED ISSUE #41: Implement automated tests
A few remaining issues are waiting on getting an automated testing framework set up (#18 #27). 

We do run one test in CI currently:

```python
import topotoolbox as topo

dem = topo.gen_random()
assert (dem.fillsinks() >= dem).z.all()
```

which I added more or less to ensure that everything is built and linked properly. This is run on the command line with `python -c`, so we could start with turning that script into a proper test.

We should at least run all the other public API functions.

The test `dem.fillsinks() >= dem` is a property-based test that we also check in the libtopotoolbox tests. Those tests in libtopotoolbox should catch errors in the libtopotoolbox implementation,  but it might be useful to run some of them in pytopotoolbox to make sure that we're passing arguments and managing the results correctly. The libtopotoolbox tests are quite messy, and it might be hard to figure out what they are testing, so I'll work on cleaning them up and documenting the properties more explicitly.

### LINKED ISSUE #50: Row/column major problem with fillsinks
While looking at #48, I noticed that `fillsinks` seems to give the wrong answer when you pass it a DEM that isn't square:

```python
import topotoolbox as topo
import matplotlib.pyplot as plt
import numpy as np

dem = topo.load_dem("tibet");
filled = dem.fillsinks();

fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12,4))
ax1.imshow(dem,cmap='terrain')
ax1.set_title('original')
ax2.imshow(filled,cmap='terrain')
ax2.set_title('filled')
plt.show()
```
filled_dem

This looks like we're passing the row-major Numpy array to libtopotoolbox in the wrong order. If I swap the rows and columns arguments to `grid_fillsinks`, I get something more like what I would expect:

```python
output = np.zeros_like(dem.z)
topo._grid.grid_fillsinks(output,dem.z,dem.columns,dem.rows)
plt.imshow(output,cmap='terrain')
```
expected

I investigated a little bit, and I think this may not have worked properly since we added the pybind11 interface to `fillsinks`. I remember swapping the array order when we were using `ctypes` (https://github.com/TopoToolbox/pytopotoolbox/pull/8#discussion_r1555717678), but it looks like we didn't copy that behavior properly when we added the `fillsinks` binding in #14. 

I am a little confused because I could have sworn I had looked at the output of `fillsinks` since we merged #14. Have you noticed this recently, @Teschl? I do want to make sure that this hasn't arisen because of a more recent change or something that changed in libtopotoolbox.

In any case, this should be an easy fix: swap the order of `dem.columns` and `dem.rows` when you pass them to `grid_fillsinks`. It would be nice to have a test (#41) that captures this in case it breaks again in the future -- I am thinking ahead to TopoToolbox/libtopotoolbox#68, in particular.

We also need to change this behavior in `identifyflats`, which seems like it should have the same problem.

Teschl:
In this pull request, I have fixed the problem of row-major/column-major when passing the matrix. Much like in @bgailleton  's solution, I'm using the `order='F'` argument.
```python
grid.z = dataset.read(1).astype(np.float32, order='F')
```
Initially, I experimented with transposing the matrix, but just using `my_matrix.T` does not solve the row-major/column-major problem. I don't think we need to rotate our matrices in order for the functions to work, right?
If not, this fixes #50 .

Other than that, I added a few basic tests and updated the ci.yaml accordingly. 
closes #41 

The oldest supported version is python 3.10 right now. We should figure out if we also want to make it compatible with older versions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,10 +14,10 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,24 +4,12 @@ on:
   push:
     branches: ["matrix-passing-fix"]
 jobs:
-  build:
+  build-and-test:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-    steps:
-      - name: Checkout package
-        uses: actions/checkout@v4
-      - name: Setup Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.10'
-      - name: Install package
-        run: python -m pip install .[opensimplex]
-        
-  test:
-    runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
@@ -29,7 +17,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: '3.10'
+          python-version: ['3.9', '3.10', '3.11']
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,10 +20,11 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-          
+
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          pip install pytest
           pip install .[opensimplex]
 
       - name: Run tests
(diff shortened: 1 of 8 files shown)


wkearn (approved):
Excellent work, @Teschl, especially with getting the testing infrastructure set up. This will be very useful going forward.

wkearn:
This fix from @Teschl works and comes with tests, so I've merged it. 

I am somewhat concerned about the performance implications of reordering arrays. With the current code, since reading a GeoTIFF loads the DEM into column-major order which is then maintained throughout the various TopoToolbox functions, I think we won't see too many performance problems. But if someone needs to integrate TopoToolbox with tools that expect arrays in row-major format --which is more natural for Python users -- they might have to reorder their data across that interface.

It should be possible to leave the data in its native layout and pass the array dimensions to libtopotoolbox in the opposite order. The Python interface would check which order the DEM is stored in and if it is `F` call `grid_fillsinks(output, dem, self.rows, self.columns)`, or if it is `C`, call `grid_fillsinks(output, dem, self.columns, self.rows)`. 

Eventually we will have a more general interface for strided array access in libtopotoolbox (TopoToolbox/libtopotoolbox#23), which will require some changes to this interface but which should then allow users to pass in general Numpy arrays (https://numpy.org/doc/stable/dev/internals.html) without worrying too much about these issues.

Issue #53, opened 2024-07-24
Pass dimensions as an array


wkearn:
I merged TopoToolbox/libtopotoolbox#69, which changes the libtopotoolbox API to take a `ptrdiff_t dims[2]` instead of two separate arguments `nrows` and `ncols`. We should update the pytopotoolbox bindings accordingly.

For row-major arrays, one would create `ptrdiff_t dims[2] = {ncols,nrows}`, while for column-major arrays one uses `ptrdiff_t dims[2] = {nrows,ncols}`. 

Ideally we would be able to pass the dimensions in the correct order and then skip the reordering steps that we introduced in #51. 

I am not sure if it makes more sense to create the `dims` object in Python and pass it to the C++ binding or if we should pass the dimensions separately along with some kind of row/column major flag to the C++ binding and organize the `dims` array there.

This change will show up in the 2024-W31 release of libtopotoolbox on 29-07-2024.

PR #54, opened 2024-07-26
Add Excesstopography

### LINKED ISSUE #42: Add excesstopography function

Teschl:
In this pull request, the excesstopography function is added to the topotoolbox. I also added some missing type hints to other functions.

Functionality:
- supports single floats/ints, ndarrays or GridObjects as arguments for threshold_slopes.
- both `fsm2d` and `fmm2d` are available
- will automatically convert arrays/GridObjects to use right datatype/order
- an example notebook has been added to the example gallery

closes #42 

Missing: (for later pull requests)
-  have only added one test to this function as of right now. I think we should use one of our future meetings to discuss what we want to test exactly (in each function). It's very expensive to run the same test that are run for the libtopotoolbox again in python.
- like we discussed, `fmm3d` is not implemented
- All of this is not supporting the new `dims` in TopoToolbox/libtopotoolbox#69 and #53


Teschl:
> It's very expensive to run the same test that are run for the libtopotoolbox again in python.

To add to that, the windows CI pipeline that builds and test already takes 2 minutes on Windows. I can't say how much of that is building, but I assume adding more (expensive) test will increase the needed time a lot.
I don't know what duration is to be expected, but it does take drastically longer than mac/Ubuntu.

wkearn:
> To add to that, the windows CI pipeline that builds and test already takes 2 minutes on Windows. I can't say how much of that is building, but I assume adding more (expensive) tests will increase the needed time a lot. I don't know what duration is to be expected, but it does take drastically longer than mac/Ubuntu.

The CI time is because libtopotoolbox takes forever to build on Windows, which I have also noticed in the libtopotoolbox CI. I don't really know what's going on there: it seems like it spends most of the time configuring the build rather than compiling the library. I'll open an issue over there to remind me to look into it more closely

wkearn:
>     * have only added one test to this function as of right now. I think we should use one of our future meetings to discuss what we want to test exactly (in each function). It's very expensive to run the same test that are run for the libtopotoolbox again in python.

I'm with you on not retesting all of the libtopotoolbox functionality: #55

wkearn (approved):

PR #58, opened 2024-08-06
Rework example gallery

### LINKED ISSUE #18: Update README.md
Some needed improvements:

- [x] Better explain how to build the package from the repository, especially going into more detail for installing on Windows
- [x] Remove outdated information
- [x] Include a short explanation on where to find the examples and what to do with them
- [x] adding badges
- [ ] how to run tests (needs tests to be added #9 )

This list is not complete at all, just to get the README.md up to date.

### LINKED ISSUE #27: Set up CI through GitHub Actions
At the moment, we mainly want to verify that the package builds correctly on different operating systems.

I don't think we really have automated tests yet, but once we have those, we can run them through CI as well. 

This is related to

- #26 
- #25 
- #23

### LINKED ISSUE #30: Set up documentation
We have had some discussion at TopoToolbox/libtopotoolbox#39 about documentation for the Python library.

Some tasks:

- [x] #9
- [x] #18 
- [x] #31
- [x] #27 
- [x] #32

### LINKED ISSUE #36: Add more context to example notebooks
The example notebooks in #33 don't have a lot of information about what they are demonstrating. I imagine some of these notebooks might get incorporated into #32, but any that remain on their own could do with a bit more text, more descriptive titles and thumbnails, etc.

### LINKED ISSUE #46: Bug: Sphinx docs grid_object.__init__ added twice
In the docs, the recursive autosummary includes the __init__.py function twice. Since that part is auto generated, I don't have a way to fix this jet.

Teschl:
This reworks the example gallery in the documentation. It's possible to add custom thumbnails to each tile now. I have reduced the notebooks down to four files. Once we decide how we want to design the thumbnails, we can close 36. I also added a reference to the tests in our README and resolved #46 (`__init__` twice in docs).

I spend way too long trying to find this solution. Since nbsphinx is not the only extension that can generate galleries, I tried to implement a few of them before I finally figured out this, admittedly, very easy solution that is implemented now. 

Seaborn for example uses [Sphinx Gallery](https://sphinx-gallery.github.io/stable/index.html), and they were able to easily add custom thumbnails. I first tried to convert our docs to support Sphinx Gallery. This requires `.py` files that will in turn be converted to `.ipynb` files when building.  That would mean either having the examples stored twice or not having any notebooks to work with. So I scrapped this approach again.

closes #18 

I think we can also close 27 since that is all implemented now. Therefor that also resolves 30


wkearn (approved):

wkearn:
Looks great, @Teschl!

I edited your PR description so it wouldn't close all of the issues immediately.

PR #59, opened 2024-08-07
Add linting and type checks

Teschl:
I added another job in the ci.yaml that contains linting and type checking. I used pylint and mypy respectively. Since the `tests/` are still very underdeveloped, I have not added them to the linting tests for now.

We need to decide how we want to handle linting issues. Right now, the result is:

> Your code has been rated at 9.92/10

Which is due to the amount of arguments in some functions. I don't think the pipeline should fail is the score is this close to 10. I'm looking into a solution for this next.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,6 +50,7 @@ jobs:
         run: | 
           pip install mypy
           pip install pylint
+          pip install flake8
           pip install numpy
           pip install matplotlib
           pip install rasterio
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,4 +62,4 @@ jobs:
       #  run: pylint src/topotoolbox/
 
       - name: Run type checks
-        run: mypy --follow-imports=skip --ignore-missing-imports src/topotoolbox
+        run: mypy --ignore-missing-imports src/topotoolbox
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,8 +25,34 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install pytest
+
+      - name: Build and install TopoToolbox
+        run: |
           pip install .[opensimplex]
 
       - name: Run tests
         run: |
           python -m pytest
+
+  lint:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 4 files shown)


Teschl:
I have changed the linter to flake8, which is a lightweight alternative to pylint. Now, the linting passes, but mypy causes issues that I can't reproduce locally.

wkearn:
I get the error that we see on CI with a fresh venv and the various packages you have installed. My guess is that means we need to pin some versions.

This is the output of `pip list` after I install everything in that venv:

```
Package           Version
----------------- -----------
affine            2.4.0
astroid           3.2.4
attrs             24.2.0
certifi           2024.7.4
click             8.1.7
click-plugins     1.1.1
cligj             0.7.2
contourpy         1.2.1
cycler            0.12.1
dill              0.3.8
flake8            7.1.1
fonttools         4.53.1
isort             5.13.2
kiwisolver        1.4.5
matplotlib        3.9.1.post1
mccabe            0.7.0
mypy              1.11.1
mypy-extensions   1.0.0
numpy             2.0.1
packaging         24.1
pillow            10.4.0
pip               22.0.2
platformdirs      4.2.2
pycodestyle       2.12.1
pyflakes          3.2.0
pylint            3.2.6
pyparsing         3.1.2
python-dateutil   2.9.0.post0
rasterio          1.3.10
setuptools        59.6.0
six               1.16.0
snuggs            1.4.7
tomli             2.0.1
tomlkit           0.13.0
typing_extensions 4.12.2
```

If I install mypy v1.11.1 and flake8 v7.1.1 and pylint v3.2.6 in my user installation, I can run `mypy` on topotoolbox successfully. The only differences in the versions between the venv and my user installation are things like `rasterio` and `numpy`, so I'm going to try to upgrade those to see when our mypy invocation breaks.

wkearn:
Having done some research it seems that this is a known issue with mypy and numpy>=2.0.0 [mypy 17396](https://www.github.com/python/mypy/issues/17396) when we use the `--follow-imports skip` flag. The discussion on that issue suggests that `--follow-imports skip` is probably not the right thing to be using.

Without it, we do get some type errors, but I fixed them up pretty easily. I'll push a couple of commits to your branch and see if everything works out on CI.

wkearn:
@Teschl, looks like CI is green now. Can you take a look at my changes to make sure that I haven't done anything weird? If everything looks okay, mark this PR as ready for review, and I can go ahead and merge it.

Teschl:
Everything looks sensible :+1: You should be able to merge it.

PR #63, opened 2024-09-03
Implement gradient8

Teschl:
This pull request includes the implementation for the gradient8 function. Before you merge it, we'll need to wait for next week's release (or which ever release contains gradient8). I will adjust the CMakeLists.txt when the new release is available.

This should be merged after #65.

PR #65, opened 2024-09-07
Add StreamObject

Teschl:
This pull request adds the StreamObject. In order for the different object to use a similar logic, some small changes have been made to the FlowObject and GridObject.

Changelog:

- Add mixins to FlowObject and StreamObject
- Change show function of FlowObject to better fit in with the other objects (only show one plot).
- Add StreamObject example
- Change C wrapper naming logic


Teschl:
I removed the show function as well as modified the magic functions in the FlowObject. This should be ready to be merged now @wkearn .

src/topotoolbox/stream_object.py
```diff
@@ -0,0 +1,78 @@
...
+        self.name = flow.name
+
+        # raster metadata
+        self.acc = acc
+        self.shape = flow.shape
+
+        # georeference
+        self.bounds = flow.bounds
+        self.transform = flow.transform
+        self.crs = flow.crs
```
wkearn, line 36: The `STREAMobj` defined in MATLAB is a so-called "node attribute list" that is an array of pixel indices that are considered part of the stream network. Typically those pixels are ones with a flow accumulation value greater than some threshold. The constructor for `STREAMobj` looks something like

```
STREAMobj(FD : FLOWobj, threshold, units)
```

where `threshold` can be either a number, in which case the stream network is the collection of pixels with a flow accumulation greater than that threshold when measured in the supplied `units` -- `flow_accumulation` returns the flow accumulation in pixels, but you could convert that to square meters or square kilometers using the `cellsize`.

`threshold` could also be a Boolean-valued `GRIDobj`, in which case the stream pixels are those where `threshold` is True.

I think we should keep the same or a similar interface for `StreamObject` here. What you have currently is a `GridObject` by another name.

src/topotoolbox/stream_object.py
```diff
@@ -0,0 +1,78 @@
...
+    """A class to represent stream flow accumulation based on a FlowObject.
+    """
+
+    def __init__(self, flow: FlowObject) -> None:
+
+        acc = np.zeros_like(flow.target, order='F', dtype=np.float32)
+        weights = np.ones_like(flow.target, order='F', dtype=np.float32)
+
+        _flow.flow_accumulation(
+            acc, flow.source, flow.direction, weights, flow.shape)
```
wkearn, line 24: `flow_accumulation` could be a method on `FlowObject` that returns a `GridObject`. It is useful on its own.

It might also make constructing a `StreamObject` simpler: the constructor really takes a `FlowObject` and a Boolean `GridObject` and extracts the stream network indices from the `GridObject` and copies the `source`, `target` and `direction` members from the `FlowObject`, which it needs to route flow along the stream network.

If the user passes a threshold value rather than a `GridObject` then we can use `flow_accumulation` followed by a comparison to construct a Boolean `GridObject` and then use the previous constructor.

src/topotoolbox/stream_object.py
```diff
@@ -0,0 +1,78 @@
...
+        Parameters
+        ----------
+        cmap : str, optional
+            Matplotlib colormap that will be used in the plot.
+        """
+        plt.imshow(self.acc, cmap=cmap)
+        plt.title(self.name)
+        plt.colorbar()
+        plt.tight_layout()
+        plt.show()
```
wkearn, line 51: It is probably fine for now to plot the accumulation or the logical `GridObject` defining the stream network (see the other comments).

The MATLAB version plots the stream network as a collection of lines. This is somewhat complicated because it requires splitting off each branch of the stream network and plotting it independently, which in turn requires traversing the graph of the stream network. It might be best to save that for future work.

src/topotoolbox/stream_object.py
wkearn: We also need to hang on to the connectivity information from the `FlowObject` so that we can traverse the stream network and do various calculations with it. MATLAB doesn't store the entire `source` and `target` arrays, but only the subsets that are part of the stream network. Find all the entries in `source` that are labeled `True` in `threshold` and then store those entries and the corresponding `target` and `direction` entries in the `StreamObject`.

Let me know if that makes sense -- I am finding it very confusing at the moment.

wkearn (changes requested):
See the comments about the `StreamObject` data structures.

src/topotoolbox/stream_object.py
Teschl: I think I understand. So `threshold` is either a `int | float` or a `GridObject`. 

1. If it's a number, we'll check if a cell is `>= threshold`. If so, it'll be part of the stream.
2. If it's a (bool) `GridObject` only the cells where the value is 1 will be part of the stream.

Only the indexes  of the stream are going to be saved in the StreamObject.
grafik
So in this case, where the blue line is the resulting stream, the array would contain the indexes like this: `[1,2,5,8]` correct?

So the `source`, `target` and `direction` arrays are also just arrays that contain the respective values when parsing through the  `[1,2,5,8]` array?

So if I continue the example (but filled with example values instead of indexes this time):
grafik
It should look like this: `source=[1,4,7,6], target=[2,1,4,5], direction=[5,5,8,6]`?

wkearn: > I think I understand. So `threshold` is either a `int | float` or a `GridObject`.
> 
>     1. If it's a number, we'll check if a cell is `>= threshold`. If so, it'll be part of the stream.
> 
>     2. If it's a (bool) `GridObject` only the cells where the value is 1 will be part of the stream.
> 

This is correct

> Only the indexes of the stream are going to be saved in the StreamObject. grafik So in this case, where the blue line is the resulting stream, the array would contain the indexes like this: `[1,2,5,8]` correct?

Yes, this is what we need.

> So the `source`, `target` and `direction` arrays are also just arrays that contain the respective values when parsing through the `[1,2,5,8]` array?
> 
> So if I continue the example (but filled with example values instead of indexes this time): grafik It should look like this: `source=[1,4,7,6], target=[2,1,4,5], direction=[5,5,8,6]`?

The `source`, `target` and `direction` arrays are "edge attribute lists" while the `stream` array is a "node attribute list." If you think of the flow network as a graph, the pixels of the raster are the nodes and edges connect two pixels. Edge attribute lists label the edges, and node attribute lists label the nodes. For now we only route flow in a single direction per pixel, so there is one edge for each node, and the lists are approximately the same length. The only potential difference depends on how you treat nodes that have no downstream neighbor (i.e. outlets or sinks). In the future we will want to support multiple flow directions, in which case there can be more edges than nodes.

In your example, if we assume that the stream flows in the direction 1 -> 2 -> 5 -> 8, then your three arrays will be

- `source = [1,2,5,8]`
- `target = [2,5,8,-1]` (the flow routing routines return a target of `-1` for any pixel that does not have a downstream neighbor).
- `direction = [4,1,1,0]` (the `direction` array encodes direction as a bitfield. These may not be the right values, because it depends on the memory order of the array).

The MATLAB implementation leaves off the `8 -> -1` edge, so it will have `source = [1,2,5]` and `target = [2,5,8]`, so you could also go through and filter out edges that have `target = -1` if you wanted to.

Teschl:
The `StreamObject` now contains 1D arrays, which only contain the data of the cells that are part of a stream. The `threshold` is set to 1 as of right now, I'm thinking we might implement a default `threshold` that is calculated based on the size of the DEM. When using the show() function, a new 2D array is generated based on the stream array. By using the `overlay` attribute, the stream can be overlaid over a DEM. The `units` attribute is something I haven't jet really implemented.

Teschl:
I added that the threshold will be automatically computed if there is no custom user input. I used the following function:

```python
shape = (n,m)
avg_shape = (n + m)//2
threshold = (avg_shape * sqrt(avg_shape))//2 + 1 
```

This function provided satisfying results in the examples I ran, but if you think we should adjust it I'm happy to do so.

wkearn:
The MATLAB version uses a default threshold of 1000 pixels, which does
not scale with the size of the DEM. My guess is that users almost always
set the threshold based on the data they have available to them, so it
might not be so important exactly what our default is. Perhaps
@wschwanghart has some more insight on the upslope area threshold for
the channels in the STREAMobj.

In any case, your formula is not dimensionally consistent. The
threshold should have units of area. If you imagine each pixel has an
edge 1 meter long, then your formula gives you units of meters^(3/2) rather
than meters^2. This might lead to unexpected behavior when you have
much larger or smaller DEMs or, for example, if you resample a DEM to a
higher or lower resolution.

wschwanghart:
Actually, I like the minimum threshold approach of @Teschl . 1000 pixels is quite arbitrary and fails (or returns an "empty" STREAMobj) if the DEM is very small. As @wkearn says, users will likely supply a value, but you never know... So, I will probably adopt @Teschl's approach.

wschwanghart:
But I agree, @wkearn . The channel initiation area should have some areal unit.

Teschl:
I added support for the actual `units` argument. I hope I understood the Matlab implementation correctly, but his should add units of area to the threshold. The following examples will result in the same StreamObject: 

```python
dem = topo.load_dem('perfectworld')
flow = topo.FlowObject(dem)

# cellsize is 100 meters in this example 
stream = topo.StreamObject(flow, units='pixels', threshold=10000)
stream = topo.StreamObject(flow, units='km2', threshold=100)
stream = topo.StreamObject(flow, units='m2', threshold=100000000)
```

+ pixels = km2 / (0.001*cellsize)^2
+ pixels = m2 / (cellsize)^2

> The threshold should have units of area.

Does this satisfy the need for a real unit of area or am I misunderstanding?

wkearn:
I think your `cell_area` calculations are correct, so if a user supplies
a threshold area in the specified units, things should work out
correctly. That might be a good test to add at some point: take a DEM
and compute the `StreamObject` with a few different thresholds in
different units, and make sure that the minimum flow accumulation for
any pixel in the stream network is greater than or equal to that area.

The problem is in your formula for the default threshold:

```python
avg = (flow.shape[0] + flow.shape[1])//2
threshold = (avg * math.sqrt(avg))//2 + 1
```

This threshold will not have the right units. `avg` has units of length
because it is the average of the lengths (in pixels) of the two sides of
the DEM.  `avg * math.sqrt(avg)` has units of length^(3/2), which
is not an area. You need to have a term like `avg * avg` instead. If you
replace `avg * math.sqrt(avg)` with `avg * avg`, though, I think the
threshold will probably be too large, so you'll need to scale it by some
fraction. I don't know off the top of my head what a good fraction might
be, but a little experimentation should be able find a good one.
Remember that there isn't a perfect default threshold and most
people will set their own. We just want to make sure that whatever we
use as a default is scientifically reasonable.

Teschl:
I changed the formula to: `(avg^2)*0.01`. 👍

src/topotoolbox/stream_object.py
```diff
@@ -0,0 +1,211 @@
...
+        temp_source = []
+        target = flow.target.flatten(order='F')
+        temp_target = []
+        direction = flow.direction.flatten(order='F')
+        temp_direction = []
+
+        for i in self.stream:
+            temp_source.append(source[i])
+            temp_target.append(target[i])
+            temp_direction.append(direction[i])
```
wkearn, line 137: Same thing here as above: building up these lists one by one is likely to be slow. But you can index with the `stream` array:

```python
source = flow.source.ravel(order='F')[stream]
```
etc.

src/topotoolbox/stream_object.py
```diff
@@ -0,0 +1,211 @@
...
+
+        self.target = np.array(temp_target, dtype=np.int32)
+        self.source = np.array(temp_source, dtype=np.int32)
+        self.direction = np.array(temp_direction, dtype=np.int32)
```
wkearn, line 141: These should be `np.int64` to be consistent with the flow object arrays.

src/topotoolbox/stream_object.py
```diff
@@ -0,0 +1,211 @@
...
+                       f"match FlowObject shape: {self.shape}.")
+                raise ValueError(err) from None
+            threshold = threshold.astype(np.float32, order='F')
+        else:
+            if threshold.shape != self.shape:
+                err = (f"Threshold GridObject shape {threshold.shape} does "
+                       f"not match FlowObject shape: {self.shape}.")
+                raise ValueError(err) from None
+
+            threshold = threshold.z.astype(np.float32, order='F')
```
wkearn, line 101: When a user passes an array to `STREAMobj` in MATLAB, it is interpreted as a Boolean array that defines the pixels that are included as part of the stream network. Imagine a user constructing a flow accumulation raster on their own and then thresholding it themselves. In Python, I imagine something like:

```python
acc = flow.flow_accumulation()
W = acc.z >= 10000
StreamObject(flow,W)
```

Maybe we use a different keyword argument for the array `W` in this case, Something like `stream_pixels`?

In the constructor of the `StreamObject`, you use this array instead of checking `acc[i] >= threshold`.

It works like this in case the user has a different definition of the stream network than the flow accumulation threshold. Maybe it comes from a digitized map or another hydrological model.

What you have implements a spatially varying threshold. I think your implementation is correct, but it does not match the MATLAB interface, which I think we should follow in this case. If a user does want to supply a spatially varying threshold, the Boolean array method described above will let them do that by computing the spatially varying threshold array on their own, and supplying `acc >= threshold` as the `W` array above.

wkearn (changes requested):
Looks pretty good, @Teschl! I think there are a few more things that we take care of now to harmonize the Python and MATLAB interfaces and to improve how we construct the `StreamObject` arrays.

I also merged #57 this morning, which causes a few merge conflicts that you'll need to resolve by rebasing/merging into your PR branch. Make sure when you finish that you are using the `2024-W40` version of libtopotoolbox and that the new `graphflood` module gets imported in `__init.py__`.

src/topotoolbox/stream_object.py
Teschl: The way this is implemented right now ignores a potential `threshold` if a `stream_pixels` is provided. Is this intended or should the threshold still be used? I assume not, since like in your example: `W = acc.z >= 10000` accounts for the threshold already.

src/topotoolbox/stream_object.py
```diff
@@ -0,0 +1,218 @@
...
+        # on stream_pixels without the need for a threshold
+        if stream_pixels is not None:
+            if stream_pixels.shape != self.shape:
+                raise ValueError(
+                    f"Stream pixels array shape {stream_pixels.shape} does "
+                    f"not match FlowObject shape {self.shape}.")
+            self.stream = np.nonzero(
+                stream_pixels.flatten(order='F'))[0].astype(
+                np.int64)
+            print(self.stream)
```
wkearn, line 93: Is this a leftover debugging statement?

src/topotoolbox/stream_object.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # Based on the stream array, generate 3 1D arrays where the value of
+        # the stream array at each index holds respective value of the
+        # original array. (source, target and direction)
+        self.source = flow.source.ravel(
+            order='F')[self.stream].astype(np.int64)
+        self.target = flow.target.ravel(
+            order='F')[self.stream].astype(np.int64)
+        self.direction = flow.direction.ravel(
+            order='F')[self.stream].astype(np.int64)
```
wkearn, line 149: The `direction` array should actually be uint8. Sorry, I think I told you the wrong type last time around.

wkearn (changes requested):
Just a couple minor things to clean up

src/topotoolbox/stream_object.py
wkearn: This is correct. If a user supplies a `stream_pixels` array, then we should ignore the threshold. I think how you have it warning the user if they pass both is reasonable.

src/topotoolbox/stream_object.py
Teschl: Yes it was. Removed now :+1:

wkearn (approved):

wkearn:
@Teschl: It looks good from my perspective. Is there anything else you wanted to add before I merge this?

Teschl:
No, I think it should be complete now. Should be good to go :+1: 

I will check #63 after this to resolve any conflicts.

PR #66, opened 2024-09-20
Fix the build time version of numpy to 2.0.0 or greater

wkearn:
NumPy 2 can build binaries that are compatible with the NumPy 1 C ABI, but not vice versa. As long as we build with NumPy 2, users should be able to use older versions of NumPy without problems.


wkearn:
According to https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-abi-handling, it might also be a good idea to run a CI test against an older version of numpy.

PR #72, opened 2024-10-30
Add anisotropic coefficient of variation

Teschl:
This adds the wrapper for the acv function to pytopotoolbox. I did not add a .ipynb file for the documentation for this function since there is not much to this function.


Teschl:
I changed the git tag to next weeks release so it's failing right now.

Teschl:
This contains the wrapper for the acv function and the new wrapping guide #76

wkearn:
Do you want to try to extract your function wrapping guide and make that a separate pull request? That would be very useful to have, and it seems like our progress on `acv` is stalled while we try to resolve its issues over in libtopotoolbox.

Teschl:
Yes that probably a good idea :+1:

PR #87, opened 2024-11-20
Update installation instructions in README and docs

wkearn:
Now that pytopotoolbox is available on PyPI (#86), installing it for users is just `pip install topotoolbox`.

@teschl: I deleted all of the previous installation instructions, but let me know if you think it would still be valuable to have some instructions on setting up a development environment.


Teschl:
I feel like keeping the instructions to build the package locally from the repository is worth keeping, mainly for windows developers. Otherwise, it might be hard to figure out yourself what `Tools for Visual Studio` are needed to compile the libtopotoolbox. 

But we could move that to a different section to avoid confusion. It's not needed in the guide for PyPi so we could hide it somewhere in the docs.

wkearn:
I copied the old installation instructions including the Windows-specific stuff to a new page of Developer Documentation and added the instructions for testing and linting and setting up the development environment there as well.

PR #96, opened 2024-11-29
Move example notebooks to docs/examples

### LINKED ISSUE #81: Rework URLs for docs
The way the docs generation is set up right now, the example notebooks are accessed with the following URL `https://topotoolbox.github.io/pytopotoolbox/_temp/`, meaning they get saved in `_temp/` which is not really a logical name when looking at the URL. I propose this should be changed to something along the lines of `examples/`

wkearn:
Fixes #81


wkearn:
@Teschl, @bgailleton: is this a good idea? In addition to cleaning up the URLs (#81), this simplifies the build process, because we don't have to copy the notebooks into the docs directory. 

On the other hand, it seems like the top-level `examples/` directory is a common pattern in Python, so I could see an argument for keeping that around.

Teschl:
I think it's fine. Especially because they are kept without outputs in the repo, they only have them on the website where the examples are easy to find anyway.

bgailleton:
Ah sorry i did not notice the tag. I do not see why it would be a problem as long as the `README.md` points to the example folder explicitly!

PR #100, opened 2024-12-06
Add Curvature to GridObject

Teschl:
This adds the curvature function to the `GridObject`. Potential issue with the use of an epsilon to avoid divisions by zero, since the Matlab version does not have this. (This branch has a lot of commits since I used it before to locally test the curvature function written in C)


wkearn (approved):
Looks pretty good, @Teschl!

Notes for future reference: Curvature is numerically unstable and very sensitive both to the data and to numerical errors. There are numerical differences between this implementation and the MATLAB one especially when the denominator of the calculation is close to zero (i.e. when the gradient is zero). We should be careful about making comparisons between the two. The MATLAB implementation sets the curvature to zero when the denominator is exactly zero. This implementation instead adds a small offset to the denominator.

PR #107, opened 2024-12-07
Add filter function to GridObject

### LINKED ISSUE #103: Add filter() function to the GridObject
As mentioned in https://github.com/TopoToolbox/libtopotoolbox/issues/103#issuecomment-2522617594.

Teschl:
This pull request adds the filter function to the `GridObject`.

closes #103


wkearn:
Merging #100 created a few merge conflicts here.

src/topotoolbox/grid_object.py
```diff
@@ -302,6 +304,78 @@ def excesstopography(
...
+                convolve(self.z, kx, mode='nearest'))
+
+        elif method in ['wiener']:
+            filtered = wiener(self.z, mysize=kernelsize)
+
+        elif method in ['std']:
+            # This solution is based on this thread:
+            # https://stackoverflow.com/questions/19518827/what-is-the-python
+            # -equivalent-for-matlabs-stdfilt-function
+            filtered = generic_filter(self.z, np.std, size=kernelsize)
```
wkearn, line 365: MATLAB's `std` and numpy's `std` divide the sum of squares by different values to compute the sample variance and standard deviation (see [Bessel's correction](https://en.wikipedia.org/wiki/Bessel's_correction) for more details). To match the MATLAB behavior, you need to multiply the output by `np.sqrt(kernelsize**2 / (kernelsize**2 - 1))`.

wkearn: This is also incredibly slow. I think it is all right for now since there doesn't seem to be a better alternative readily available in `scipy`, but it is worth investigating in the future.

src/topotoolbox/grid_object.py
```diff
@@ -302,6 +304,78 @@ def excesstopography(
...
+        elif method in ['median']:
+            filtered = median_filter(self.z, size=kernelsize, mode='reflect')
+
+        elif method in ['sobel', 'scharr']:
+            if kernelsize != 3:
+                arr = f"The method '{method}' only works with a 3x3 kernel'."
+                raise ValueError(arr) from None
+
+            if method == 'sobel':
+                ky = np.array([[1, 2, 1], [0, 0, 0], [-1, -2, -1]])
```
wkearn, line 349: There is also a `'sobel` function in `scipy.ndimage`, but it seems to give the same results as the explicit filter you use here, and I don't think there is any real benefit to using one over the other.

wkearn (changes requested):
Looks good. There is just the issue with the Bessel correction to fix.

Teschl:
I fixed the issue with the `std` method. Should be ready to be merged. @wkearn

wkearn (approved):

PR #116, opened 2024-12-10
Add erode to GridObject

### LINKED ISSUE #105: Add erode() function to GridObject
As mentioned in https://github.com/TopoToolbox/libtopotoolbox/issues/103#issuecomment-2522617594

Teschl:
This adds the `erode()` function to the `GridObject`. It wraps the `scipy.ndimage.grey_erosion()` function. Like in the Matlab wrapper, all NaNs are set to `inf` before running the erosion and reverted to NaNs afterward. 

The [Matlab](https://de.mathworks.com/help/images/ref/imerode.html#d126e161642) wrapper only supports the argument `SE` (Structuring element) and ignores `nhood`. In the [Python](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.grey_erosion.html) implementation, `footprint` is `nhood` and `structure` is `SE`. I kept both (and `size`) so provide as much customizability as possible to the user.

Right now, the function defaults to size=(3,3) if nothing else is provided. There is no specific reasoning behind that, so if you have a suggestion what the default should be, I will adjust that. @wkearn (The pipeline is failing because there is still this TODO in the file)

closes #105

### DIFF
--- a/src/topotoolbox/grid_object.py
+++ b/src/topotoolbox/grid_object.py
@@ -520,37 +520,45 @@ class GridObject():
             self, size: tuple | None = None, footprint: np.ndarray | None = None,
             structure: np.ndarray | None = None) -> 'GridObject':
         """Apply a morphological erosion operation to the GridObject. Either
-        size, footprint or structure has to be passed to this function.
-        Otherwise, a default size will be used.
+        size, footprint or structure has to be passed to this function. If
+        nothing is provided, the function will raise an error.
 
         Parameters
         ----------
         size : tuple of ints
             A tuple of ints containing the shape of the structuring element.
-            Only needed if neither footprint nor structure is provided.
-            Defaults to (3,3)
+            Only needed if neither footprint nor structure is provided. Will
+            result in a full and flat structuring element.
+            Defaults to None
         footprint : np.ndarray of ints, optional
-            A boolean array defining the footprint of the erosion operation.
+            A array defining the footprint of the erosion operation.
             Non-zero elements define the neighborhood over which the erosion
-            is applied. If None, the default is a full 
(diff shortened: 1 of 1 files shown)


src/topotoolbox/grid_object.py
```diff
@@ -516,6 +516,52 @@ def curvature(self, ctype='profc', meanfilt=False) -> 'GridObject':
...
+        GridObject
+            A GridObject storing the computed values."""
+
+        # Replace NaN values with inf
+        dem = self.z.copy()
+        dem[np.isnan(dem)] = np.inf
+
+        if size is None and (structure is None and footprint is None):
+            # TODO: choose default vaule
+            size = (3, 3)
```
wkearn, line 553: I think rather than setting a default size, we should throw an error if you don't provide any structuring element. MATLAB TopoToolbox, MATLAB's `imerode` function and scipy's `grey_erosion` all throw an error in this case. There are so many possibilities here -- different sizes, non-flat structuring elements, flat but non-zero structuring elements, structuring elements that aren't square -- that users really should specify what they mean.

This error message should be informative. Something like "erode requires a structuring element to be specified. Use the size argument for a full and flat structuring element (equivalent to a a minimum filter) or the structure and footprint arguments to specify a non-flat structuring element."

wkearn (changes requested):
Looks good. I think we should replace the default size with an error as detailed below, but otherwise I think it is ready to go.

src/topotoolbox/grid_object.py
Teschl: Done :+1:

wkearn (approved):

PR #119, opened 2024-12-19
Add Evansslope to GridObject

### LINKED ISSUE #106: Add evansslope() to GridObject
As mentioned in https://github.com/TopoToolbox/libtopotoolbox/issues/103#issuecomment-2522617594

Teschl:
This pull request adds the `evansslope` function to the `GridObject`. There are a few differences to the Matlab implementation:

- removed `removenans` argument, since `scipy.ndimage.convolve` needs NaNs to be removed (so it's always replacing NaNs with nearest neighbor)
- `padval` is renamed `mode`, to better reflect the underlying scipy function.
- Added `partial_derivitives` argument to provide the option to return fx and fy like in Matlab
- removed a bunch of code used for padding the DEM, since the padding is handled in the `convolution` function

I also had to disable the `too-many-lines` pylint error since the file is now larger than 1000 lines.

closes #106

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,4 +50,5 @@ disable = [
     "duplicate-code",
     "cyclic-import",
     "too-many-positional-arguments",
+    "too-many-lines",
 ]
--- a/src/topotoolbox/grid_object.py
+++ b/src/topotoolbox/grid_object.py
@@ -1,6 +1,7 @@
 """This module contains the GridObject class.
 """
 import copy
+from typing import Tuple
 
 import numpy as np
 import matplotlib.pyplot as plt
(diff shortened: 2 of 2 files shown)


wkearn:
> I also had to disable the `too-many-lines` pylint error since the file is now larger than 1000 lines.

I think this is fine for now, but I am starting to think that the growth of GridObject is something we need to control. I will open another issue about that, though.

wkearn (approved):
I think there are some slight differences in how NaNs get handled here, so that the results are slightly different from TopoToolbox v2 in the presence of NaNs. If/when we do eventually have something equivalent to `inpaintnans`, then we can add use that here.

PR #130, opened 2025-01-22
Plot StreamObjects with line segments (plus an API change for the plotting)

### LINKED ISSUE #111: Improve visualization of `StreamObject`
The current plotting method for `StreamObject` converts it to a binary `GridObject` and plots that. It works, but for larger DEMs, it gets very hard to see because every stream is a single pixel wide. The MATLAB `STREAMobj.plot` method splits the stream network into line segments and plots those segments. This is probably what we should do, but it does require a slightly complicated traversal of the stream network.

wkearn:
Resolves #111 

This uses `StreamObject.xy` to construct stream segments and then plots them with a LineCollection from matplotlib.

This API follows the guidelines in

https://matplotlib.org/stable/users/explain/figure/api_interfaces.html#third-party-library-data-object-interfaces

@Teschl: what do you think about doing something like this for the GridObject plotting interface as well? I think this can make our plots a lot easier to compose. A user would plot a StreamObject over a GridObject like

```python
fig, ax = plt.subplots()
dem.plot(ax=ax, cmap="terrain")
S.plot(ax=ax,color='r',linewidth=0.25)
```
The user can then have as much control over the figure and axes as they want. And by forwarding optional keyword arguments, we let users control other parts of the GridObject plot like the color range (which would be useful for the [curvature plots in the docs](https://topotoolbox.github.io/pytopotoolbox/tutorial/grid.html#Curvature)).

### DIFF
--- a/docs/tutorial/stream.ipynb
+++ b/docs/tutorial/stream.ipynb
@@ -20,6 +20,7 @@
    "outputs": [],
    "source": [
     "import topotoolbox as tt3\n",
+    "import matplotlib.pyplot as plt\n",
     "\n",
     "dem = tt3.load_dem('tibet')\n",
     "fd  = tt3.FlowObject(dem);\n",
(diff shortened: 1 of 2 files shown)


Teschl:
That seems like a great idea @wkearn . Having the `show()` was pretty restricted anyway.

We should also do something with the `show()` function provided in `utils.py`. That one can probably be removed, or replaced, as well.

Issue #113, opened 2024-12-10
Implement trunk


wkearn:
This extracts the longest stream from each connected component of a `StreamObject` and returns a new `StreamObject`. It involves a few traversals of the stream network, one to compute the downstream distance and one to keep track of the maximum distance.

This is used in the Getting Started guide section "Derive, modify and plot the stream network."

wkearn:
Good question. There are two traversals of the stream network in `trunk`. One within `distance` [here](https://github.com/TopoToolbox/topotoolbox3/blob/ce45c21a94161905cbb425fa7cf1f716a2d61bf9/toolbox/%40STREAMobj/trunk.m#L72-L76):

```
if nargin == 1
    dds = distance(S,'max_from_ch');
else
    dds = a;
end
```

and [one](https://github.com/TopoToolbox/topotoolbox3/blob/ce45c21a94161905cbb425fa7cf1f716a2d61bf9/toolbox/%40STREAMobj/trunk.m#L86-L88) that identifies the stream elements that are part of the stream with the maximum downstream distance.

```
for r = numel(S.ix):-1:1
    I(S.ix(r)) = I(S.ixc(r)) && II(S.ix(r));
end
```

We probably want to avoid computing those loops explicitly in Python. The other stuff in `trunk` extracts and renumbers the stream network nodes, and that will be easier to do in Python.

Looking into this has given me an idea for a good interface for doing these kinds of traversals in libtopotoolbox. I'll open an issue over there.

PR #147, opened 2025-02-10
Use the new libtopotoolbox flow routing API

### LINKED ISSUE #146: Adapt to flow routing API changes in libtopotoolbox
TopoToolbox/libtopotoolbox#167 introduced some changes to the flow routing API that will break things here. The main change is to `flow_routing_targets`, which became `flow_routing_d8_edgelist` and now accepts an additional parameter. Renaming and fixing this interface is fairly straightforward, but we also need to distinguish between the topologically sorted list of nodes and the source vertices for the edge list on the Python side of things. The node list contains every node in the FlowObject, which is every node in the DEM. The edge list now only contains valid edges: there are no placeholder edges from sinks/outlets to the fictitious -1 node. Functions like `drainagebasins` and `flow_accumulation` now expect the edge list rather than the topologically sorted list of nodes.

wkearn:
This should resolve #146. It required changing `flow_routing_targets` to `flow_routing_d8_edgelist` as expected, but it also required us to switch from using libtopotoolbox's `flow_accumulation` to `flow_accumulation_edgelist`. I had hoped to make this switch otherwise, so this was a good opportunity to do it. There was one other minor libtopotoolbox change to `drainagebasins` that this effort revealed (TopoToolbox/libtopotoolbox#169).

The changes to StreamObject were minimal, updating it to use the new flow accumulation interface and cleaning up some indices in the constructor function. The normal StreamObject functions already use proper edge lists, so once the constructor was fixed, everything worked properly.

The tests did catch a lot of the necessary changes. I added an index check to `test_flow_object.py` make sure that we don't include the -1 edges in the `FlowObject` edge list.

@Teschl: I don't think this will affect anything you have been currently working on (like #113), but take a look and let me know if anything seems off. As I said, the `StreamObject` internals should remain similar to what it was before because we were already creating and using the edge lists properly. It's really the `FlowObject` that has changed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ include(FetchContent)
 FetchContent_Declare(
   topotoolbox
   GIT_REPOSITORY https://github.com/TopoToolbox/libtopotoolbox.git
-  GIT_TAG 2615227a1fc0b1d2f9debc749027d720d4c6d6a6
+  GIT_TAG 2025-W06
 )
 FetchContent_MakeAvailable(topotoolbox)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ include(FetchContent)
 FetchContent_Declare(
   topotoolbox
   GIT_REPOSITORY https://github.com/TopoToolbox/libtopotoolbox.git
-  GIT_TAG 2025-W06
+  GIT_TAG 3458418513a5a92061c18a12bc02659de9ef1667
 )
 FetchContent_MakeAvailable(topotoolbox)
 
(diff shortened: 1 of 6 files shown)


Teschl:
Looks good as far as I can tell :+1: 

One thing to note is switching from `np.zeros_like` (or `np.ones_like`) to just `np.zeos`. Should we implement this change everywhere, or is there no deeper reason for it?

wkearn:
I used `np.zeros` rather than `np.zeros_like` because the `source` array is no longer the same size and shape as the original `GridObject`. So here for example:


```diff
-        acc = np.zeros_like(self.source, dtype=np.float32, order='F')
+        acc = np.zeros(self.shape, dtype=np.float32, order='F')
```

`self.source` is no longer the right size, so I used `np.zeros` with the `self.shape` parameter. I guess we could reference `direction`, which should be the same size as the DEM, but it has a different type anyway, so there doesn't seem to be any real benefit over doing `np.zeros` like this.

I think I got all of the instances where we were creating arrays based on `source`, but we should change that if we find any that I missed. Otherwise, I think it is fine to use `zeros_like` where it applies.

PR #153, opened 2025-02-12
Add trunk to StreamObject

### LINKED ISSUE #113: Implement trunk

Teschl:
This implements the [`trunk` function](https://github.com/TopoToolbox/topotoolbox3/blob/main/toolbox/%40STREAMobj/trunk.m). I (finally) got the trunk function working, more or less copying the Matlab version directly but modifying it to work in the python version.

This still needs work:
- The variable names are taken directly from the Matlab version. I'm not a big fan of these variable names, they should be more descriptive. (also, they don't conform to pep8 in some places, so a failing pipeline is expected)
- It also doesn't support the use of arguments `A` (GRIDobj with flow accumulation values) and  `a` (node-attribute list). Should we include this functionality (in parts or all of it)?

@wkearn Are there variable names that should be kept in the current form, or can I go ahead and replace them by more descriptive names? I assume some of them are more or less standard for what they represent.

closes #113


wkearn:
> This implements the [`trunk` function](https://github.com/TopoToolbox/topotoolbox3/blob/main/toolbox/%40STREAMobj/trunk.m). I (finally) got the trunk function working, more or less copying the Matlab version directly but modifying it to work in the python version.

Looks good, @Teschl! I am curious if we could do something to avoid the sparse matrix manipulation, but it makes sense to have a working version based on the MATLAB implementation that we can experiment with in the future.

> This still needs work:
> 
>     * The variable names are taken directly from the Matlab version. I'm not a big fan of these variable names, they should be more descriptive. (also, they don't conform to pep8 in some places, so a failing pipeline is expected)
> @wkearn Are there variable names that should be kept in the current form, or can I go ahead and replace them by more descriptive names? I assume some of them are more or less standard for what they represent.

I think we can replace these variable names with more descriptive ones. `OUTLET` is the only scientifically meaningful one, users won't see any of this anyway, and we might as well follow the standard. I might suggest `OUTLET => outlets`, `II => max_neighbor` and `I =>  trunks`. Feel free to use ones that make more sense to you if you can think of some.


>     * It also doesn't support the use of arguments `A` (GRIDobj with flow accumulation values) and  `a` (node-attribute list). Should we include this functionality (in parts or all of it)?

This shouldn't be too hard to add because we have `StreamObject.ezgetnal`. You should be able to do something like

```python
def trunk(self, a=None):
  # ...
  if a is None:
    dds = self.downstream_distance()
  else:
    dds = self.ezgetnal(a)
```

`ezgetnal` will throw an error if `a` is not the right data type or size.

> 
> closes #113

PR #158, opened 2025-02-28
Add Prominence to GridObject

### LINKED ISSUE #157: Add prominence to GridObject
Create a wrapper for the libtopotoolbox implementation TopoToolbox/libtopotoolbox#114.

Teschl:
~~This implements the wrapper around the TopoToolbox/libtopotoolbox#178 function. The created pybind11 wrapper returns a tuple instead of just modifying data in place.  The ndarrays are created from result data using custom delete capsules to avoid deleting them early.~~

These capsules can be reworked I think, something like this should work too, but i'll wait for your opinion @wkearn 
```cpp
auto py_result_values = py::array_t<float>(
        {result_size},
        {sizeof(float)},
        result_values,
        py::cast<py::none>()
);
```
instead of
```cpp
auto py_result_values = py::array_t<float>(
        {result_size},
        {sizeof(float)},
        result_values,
        py::capsule(result_values, [](void *p) { std::free(p); })
);
```

Merging this will have to wait on the TopoToolbox/libtopotoolbox#178 pull request.

closes #157

### DIFF
--- a/src/topotoolbox/grid_object.py
+++ b/src/topotoolbox/grid_object.py
@@ -740,7 +740,7 @@ class GridObject():
         result.z = aspect
         return result
 
-    def prominence(self, tolerance: float) -> Tuple:
+    def prominence(self, tolerance: float, use_hybrid=True) -> Tuple:
         """This function calculates the prominence of peaks in a DEM. The
         prominence is the minimal amount one would need to descend from a peak
         before being able to ascend to a higher peak. The function uses image
@@ -754,6 +754,8 @@ class GridObject():
         tolerance : float
             The minimum tolerance for the second to last found peak. (meters)
             Will always find one peak.
+        use_hybrid : bool, optional
+            If True, use the hybrid reconstruction algorithm. Defaults to True.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


Teschl:
I rewrote the function in python, this requires the reconstruct function though, so that will be added with TopoToolbox/libtopotoolbox#178. Other than that, this should be complete @wkearn

wkearn:
I get errors when I try to build pytopotoolbox with your `prominence` branch of libtopotoolbox but not when I build with the recently merged commit. I think your branch is missing some recent changes to libtopotoolbox that you need to merge/rebase (or use the new commit on libtopotoolbox `main` that I mentioned above).

src/topotoolbox/grid_object.py
```diff
@@ -740,6 +740,49 @@ def aspect(self, classify: bool = False) -> 'GridObject':
...
+        indices = []
+
+        while not prominence or prominence[-1] > tolerance:
+            diff = dem - p
+            prominence.append(np.max(diff))
+            indices.append(np.unravel_index(np.argmax(diff), self.shape))
+
+            p[indices[-1]] = dem[indices[-1]]
+
+            _morphology.reconstruct(p, dem, self.shape)
```
wkearn, line 778: Could we use `reconstruct_hybrid` here by default, and let users choose one or the other, like we do in `fillsinks`? `reconstruct_hybrid` is usually faster.

wkearn (changes requested):
Other than making it possible to use `reconstruct_hybrid`, I think this looks pretty good.

src/topotoolbox/grid_object.py
Teschl: I added the hybrid version :+1:

wkearn (approved):

PR #162, opened 2025-03-07
Remove unused tests/test_utils.py and tests/__init__.py

wkearn:
@Teschl: these files don't seem to do anything. Is there a reason to keep them around? I gather [from here](https://docs.pytest.org/en/stable/explanation/goodpractices.html) that there are circumstances when you might need an `__init__.py` in the `tests/` directory, but it does not seem like that advice applies to us at the moment.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,2 +1,0 @@
-import pytest
-import topotoolbox.utils


Teschl:
I'm not sure if that had any reason to exist back when I first added very basic tests. I can't find a purpose for it right now, so I assume it can be removed safely. @wkearn

PR #192, opened 2025-03-22
fix: replace duplicate code with direct subgraph call

### LINKED ISSUE #181: Replace instances of subgraph
Code equivalent to `subgraph` is used to extract subsets in `trunk` and `klargestconncomps`. That code can be removed and replaced with a call to `subgraph`.

XueSongTap:
fix: replace duplicate code with direct subgraph calls

Replace the duplicate implementation of subgraph functionality in the
trunk and klargestconncomps methods with direct calls to the subgraph method.
This reduces code duplication, improves maintainability, and ensures
consistent behavior across all methods that extract subgraphs.

Closes #181

### DIFF
--- a/src/topotoolbox/stream_object.py
+++ b/src/topotoolbox/stream_object.py
@@ -601,18 +601,7 @@ class StreamObject():
             trunks[self.source[r]] = trunks[self.target[r]
                                             ] and max_neighbor[self.source[r]]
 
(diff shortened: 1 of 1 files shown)


XueSongTap:
self test```

wkearn (approved):
Looks great, @XueSongTap! Thanks for the contribution!

Issue #120, opened 2024-12-20
Should we split up the large classes?


wkearn:
# Problem

`GridObject` has surpassed 1000 lines of code, triggering a pylint warning. I am not so worried about the number of lines, but the file is getting somewhat hard to work with. We have not been focused on adding functionality for `FlowObject` and `StreamObject`, but they will also grow in the future.

# Context

Our architecture generally follows the MATLAB TopoToolbox v2 architecture, which has a few classes each with lots of methods. MATLAB supports this style by allowing methods to be defined in separate files. We don't use very many object-oriented features like inheritance or sophisticated abstraction techniques -- each class is largely a bag of functions that operate on that class. One feature that is particularly useful, however, is operator overloading, which lets us apply built-in operators like `+`, `-`, `<` to `GRIDobj`s.

Originally, pytopotoolbox used mixins, parent classes of `GridObject` with methods but without any data. We had a `FillsinksMixin`, and `IdentifyflatsMixin`, and some others, which implemented the respective functions, and then had `GridObject` inherit from those classes, which gave it access to the implementation. We switched from that to the current architecture in #39,  based on @Teschl's findings (#34) that mixins complicated the documentation and that having large classes is not uncommon in Python packages.

# Possible options

1. Keep what we have, work on organizing `GridObject.py` better so that it makes more sense.
2. Return to the mixin strategy.
3. Define functions like `fillsinks` as standalone functions that take `GridObject`s (see `graphflood.py` for an example).
4. Create an interface (an [abstract base class](https://docs.python.org/3/library/abc.html)?) for `GridLike` objects that includes methods like `data()` to return an array with the data values and `cellsize()` to return the horizontal resolution. Implement processing functions as standalone functions using this interface, subclass `GridObject` from `GridLike` and define `data` and `cellsize` appropriately.

What other options might help us keep our classes manageable?

PR #194, opened 2025-03-25
Implement imposemin

wkearn:
`imposemin` can take either a `FlowObject` or a `StreamObject`, so it is implemented as a standalone function in the new file src/topotoolbox/stream_functions.py. It uses the (min,+) traversal recently added to libtopotoolbox.

`imposemin` for a `FlowObject` requires the interpixel distances. I have added a method very similar to `StreamObject.distance` that computes these.

Tests are added for both the `StreamObject` and `FlowObject` versions. The tests ensure:

1. `imposemin` does not modify the original DEM. Certain copies are made explicitly in the implementation of `imposemin` to make sure that this is true.
2. The carved DEM is not above the original
3. The gradient is constrained within numerical error by the `minimum_slope` argument.

@Teschl: would you mind taking a look at this? I've done this one a little differently, using a standalone function in a new submodule instead of implementing separate but identical methods for both `StreamObject` and `FlowObject`, and I want to get some feedback on that approach before merging.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ include(FetchContent)
 FetchContent_Declare(
   topotoolbox
   GIT_REPOSITORY https://github.com/TopoToolbox/libtopotoolbox.git
-  GIT_TAG f20d66f8509eab02cc062e95ddd1ef1dbd233324
+  GIT_TAG 6b631471b31d1cfef14193087ffe0a7a20bc1bcb
 )
 FetchContent_MakeAvailable(topotoolbox)
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -8,6 +8,7 @@ API Documentation
    topotoolbox.GridObject
    topotoolbox.FlowObject
    topotoolbox.StreamObject
+   topotoolbox.stream_functions
    topotoolbox.run_graphflood
    topotoolbox.read_tif
    topotoolbox.write_tif
--- a/src/lib/stream.cpp
+++ b/src/lib/stream.cpp
@@ -99,6 +99,22 @@ void wrap_traverse_down_f32_max_add(py::array_t<float> output,
                             target_ptr, edge_count);
 }
 
+void wrap_traverse_down_f32_min_add(py::array_t<float> output,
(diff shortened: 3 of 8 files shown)


Teschl:
It might be possible to implement this like we implemented the mixin functions originally. The only issue with this might be that it would result in the use of circular imports again. If that turns out to be an issue, it should be good to use the solution you implemented. If we figure out a good solution for this, we could also use it for #120. @wkearn 

```python
# stream_functions.py
from . import StreamObject, FlowObject
class StreamFunctions():
    def imposemin(self,  ....):
        if isinstance(self, StreamObject):
            # ...
        if isinstance(self, FlowObject):
            # ...

# flow_object.py
from . import StreamFunctions
# ...
class FlowObject(StreamFunctions):
    # ...

# stream_object.py
from . import StreamFunctions
# ...
class StreamObjectObject(StreamFunctions):
    # ...
```

wkearn:
Reading more about this has convinced me that `isinstance` is often a sign that we are [doing something wrong](http://canonical.org/~kragen/isinstance/). In most cases, I think we should depend on the implicit interfaces that our objects define using duck typing.  `imposemin` would look like:

```python
def imposemin(s, dem, minimum_slope=0.0):
    # Implement FlowObject.ezgetnal, so that it returns an array 
    # when given an array but a GridObject when given a GridObject. 
    # StreamObject.ezgetnal works as it currently does (always returning an array)
    result = copy.deepcopy(s.ezgetnal(dem))

    # Implement GridObject.view to return a view into the data. 
    # This should just returns result.z.view(), so it seems unnecessary, 
    # but it would eliminate having to check whether result is a GridObject or an array.
    z = result.view()

    d = -minimum_slope * s.distance()
    _stream.traverse_down_f32_min_add(z, d, s.source, s.target)
    return result 
```

which is way nicer than what I currently have here. The `GridObject.view` method could be used in a lot of places where we check 

```python
if isinstance(dem, GridObject):
  z = dem.z
else:
  z = dem
```

and `FlowObject.ezgetnal` (maybe we could come up with a better name) would be a good place to implement the alignment checks that we do informally everywhere.

PR #196, opened 2025-03-27
Pass dtype and copy arguments to GridObject.__array__

wkearn:
These arguments are required from Numpy 2.0. This should allow us to use np.array and np.asarray to extract data from GridObjects and ndarrays identically. Care must be taken with earlier versions of Numpy, because the semantics of copying have changed.


wkearn:
CI failures are indeed due to issues with older versions of Numpy.

PR #197, opened 2025-03-27
Few improvements on GridObject

bgailleton:
This is a small pull request:

- Vectorise the element-wise comparisons and logical operators
- Add a `__repr__` function mimicking the `info` (when the object is called within the interpreter, it helps for debugging in Jupyter lab)
- Add a `duplicate_with_new_data` function to make an easy copy a GridObject with a new z but the same metadata

It passes `pylint` and `mypy`

Let me know if you need other info/changes

### DIFF
--- a/src/topotoolbox/grid_object.py
+++ b/src/topotoolbox/grid_object.py
@@ -1332,7 +1332,7 @@ class GridObject():
             raise ValueError("Invalid cell value. 'and' can only compare True (1) and False (0) values.")
 
         # Perform element-wise bitwise AND operation
-        dem.z = self.z & other.z
+        dem.z = np.logical_and(self.z, other.z)
 
         return dem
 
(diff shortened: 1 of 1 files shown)


src/topotoolbox/grid_object.py
```diff
@@ -1162,6 +1162,40 @@ def shufflelabel(self, seed=None):
...
+        -------
+        GridObject
+          A grid identical to the input, but with new data.
+
+        """
+
+        rows, columns = data.shape
+
+        if data.flags.f_contiguous:
+            columns, rows = data.shape
```
wkearn, line 1188: I think this is currently incorrect. The row/column-major distinction is a huge problem in pytopotoolbox at the moment, and we really need to sort this out ASAP. I'll open an issue to discuss this stuff more generally.

For this PR the problem is that

```python
dem.duplicate_with_new_data(np.array(dem.z))
```

fails with an error when `dem.z.flags.f_contiguous==True` because the `rows` and `columns` arguments get flipped. If you get rid of the `if data.flags.f_contiguous` check here, then I think things should mostly work out.

However, if you are then running into problems when you try to pass these arrays to other pytopotoolbox functions, especially graphflood,  then the problem lies with our inconsistent handling of memory layouts, and that is a bigger issue that we need to look into.

bgailleton: ok I tried with a couple of arrays of different orders and it seems to hold. I'll run the lint tests and commit

PR #204, opened 2025-03-28
Add a test running `excesstopography` with different memory layouts

wkearn:
See #199


wkearn:
This test should almost certainly have failed. There is a problem with using the randomly generated DEMs with `excesstopography`.

PR #208, opened 2025-03-29
Add pre-commit support and examples to docstrings

### LINKED ISSUE #179: Add Examples to Docstrings

### LINKED ISSUE #180: Add shell script to automatically run pylint, mypy, etc.
- [ ] Check for notebook metadata in ci pipeline
- [ ] Add shell script to automatically clean notebooks, run pylint, mypy etc.

Teschl:
This adds example sections to all functions to help new users get started.  These examples are basic, but showcase how to get each function to work and how to visualize their output.

There are a few functions still missing, where I wasn't sure how to create a fitting example:

- `downstream_distance` (FlowObject)
- `ezgetnal` (FlowObject)
- `xy` (FlowObject)
- `plot` (FlowObject)
- `plotdz` (FlowObject)
- `plot` (GridObject)
- `plot_hs` (GridObject)
- `shufflelabel` (GridObject)

I think just not having an example would be fine in some cases, but I'm interested in what you think. @wkearn 

I also added a `.pre-commit-config.yaml` to support the use of pre-commit. Apart from running `pylint` and `mypy` this also checks for trailing white spaces for example. I also added `nbstripout` to the config to clean all metadata from notebook files before committing them.

Note: We could add or remove different 'basic' hooks like `requirements-txt-fixer` from [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks)

closes #180 
closes #179

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.10', '3.11']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
 
     steps:
       - name: Checkout repository
@@ -35,6 +35,7 @@ jobs:
           python -m pytest
 
       - name: Downgrade numpy
+        if: ${{ matrix.python-version <= '3.11' }}
         run: |
           pip install matplotlib scipy rasterio "numpy==1.23.5" --upgrade --upgrade-strategy eager
           pytest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest, macos-13]
         cibw_arch: ["native"]
-        cibw_build: ["cp310-* cp311-* cp312-*"]
+        cibw_build: ["cp310-* cp311-* cp312-* cp313-*"]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -13,7 +13,7 @@ API Documentation
    topotoolbox.read_tif
    topotoolbox.write_tif
    topotoolbox.load_dem
-   topotoolbox.load_open_topography
+   topotoolbox.load_opentopography
    topotoolbox.get_dem_names
    topotoolbox.get_cache_contents
    topotoolbox.read_from_cache
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -47,6 +47,12 @@ pull request to pytopotoolbox, because failing lints or type checks
 will cause a test failure that must be fixed before your contribution
 can be accepted.
 
+If you run into problems setting up development environment on any
+platform, please `open an issue
+<https://github.com/TopoToolbox/pytopotoolbox/issues/new>`_. While we
+test TopoToolbox extensively on different platforms, we only develop
+on a few, so feedback on how we can make this process as smooth as
+possible is greatly appreciated.
 
 Development environment on Windows
 ----------------------------------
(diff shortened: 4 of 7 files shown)


wkearn:
I like this approach for the pre-commit hooks. 

I think it is all right if some of the functions don't necessarily have examples yet. That's a good task for new contributors.

I followed the instructions to install pre-commit, made some changes to `grid_object.py` and ran the pre-commit hook, and I get these errors for pylint

```
************* Module topotoolbox.grid_object
src/topotoolbox/grid_object.py:1072:9: W0511: TODO: Add Example to Docstring (fixme)
src/topotoolbox/grid_object.py:1153:9: W0511: TODO: Add Example to Docstring (fixme)
src/topotoolbox/grid_object.py:6:0: E0401: Unable to import 'numpy' (import-error)
src/topotoolbox/grid_object.py:8:0: E0401: Unable to import 'matplotlib.pyplot' (import-error)
src/topotoolbox/grid_object.py:9:0: E0401: Unable to import 'matplotlib' (import-error)
src/topotoolbox/grid_object.py:11:0: E0401: Unable to import 'scipy.ndimage' (import-error)
src/topotoolbox/grid_object.py:19:0: E0401: Unable to import 'scipy.signal' (import-error)
src/topotoolbox/grid_object.py:21:0: E0401: Unable to import 'rasterio' (import-error)
src/topotoolbox/grid_object.py:22:0: E0401: Unable to import 'rasterio.warp' (import-error)
src/topotoolbox/grid_object.py:23:0: E0401: Unable to import 'rasterio.enums' (import-error)
```

I think this is because pre-commit runs in an isolated virtual environment with no dependencies. Is there a good way to install our dependencies before running the hook?

From mypy I get:

```
src/topotoolbox/__init__.py: error: Duplicate module named "topotoolbox" (also at "src/topotoolbox/__init__.py")
src/topotoolbox/__init__.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
src/topotoolbox/__init__.py: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
```

Teschl:
I have added examples to all functions where I was able to create one that is short but still properly displays the use/usecase. Some functions, like `StreamObject.xy()` for example, don't have an example because I was not able to create an example that properly showcases the function. I don't think just printing the lists of (x,y) pairs is a good example, but I also wasn't able to think of an example where something is being plotted. 

Since nothing really depends on these examples, I think at this point i would be fine to add them as they are now and if the need Arieses add other examples later.

The precommit hooks should be working now. Running them the first might take a bit longer since it will install all needed packages for pylint/mypy.

docs/tutorial/flow.ipynb
```diff
@@ -68,29 +69,12 @@
...
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
-   "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.11.2"
-  }
- },
+ "metadata": {},
```
wkearn, line 77: When I modify a notebook and then run the `nbstripout` hook when trying to make a commit, it leaves the `kernelspec` and `language_info` sections in. Do we need to add some `extra-keys` to the filter?

```diff
modified   .pre-commit-config.yaml
@@ -10,6 +10,7 @@ repos:
     rev: 0.8.1
     hooks:
       - id: nbstripout
+        args: ['--extra-keys=metadata.kernelspec metadata.language_info']
   - repo: https://github.com/pycqa/pylint
     rev: v3.1.0
     hooks:
```

seems to work for me. Do you have these options set up in your git config somewhere that strips these metadata sections without the argument in `.pre-commit-config.yaml`?

src/topotoolbox/grid_object.py
```diff
@@ -391,6 +420,12 @@ def filter(self, method: str = 'mean', kernelsize: int = 3) -> 'GridObject':
         ValueError
             If the kernelsize does not match the requirements of this function
             or the selected method is not implemented in the function.
+
+        Example
+        -------
+        >>> dem = topotoolbox.load_dem('perfectworld')
+        >>> sharr = dem.filter(method='scharr', kernelsize=3)
+        >>> plt.schatt.plot(cmap='terrain')
```
wkearn, line 428: This should be `sharr.plot(cmap='terrain')`, right?

wkearn (changes requested):
Excellent work with the examples, and I think the pre-commit hooks will be very useful. I just had one question about how the `nbstripout` hook works, and one minor error in an example.

I think there are also merge conflicts in `load_opentopography` because I merged #239 first.

src/topotoolbox/grid_object.py
Teschl: Yes :+1:

Teschl: Yes :+1:

docs/tutorial/flow.ipynb
Teschl: I was thinking it would be reasonable to keep the kernel spec. When I run a notebook, this is what it looks like:

```ipynb
"metadata": {
  "kernelspec": {
   "display_name": ".venv",
   "language": "python",
   "name": "python3"
  }
 },
```

The only part that should be removed is the display_name, since it will probably be different for every user. But it also has no effect (other than being the name used in the UI). 
The other parts however prevent that you have to select your kernel every time.

BUT the display name will be used in the UI even when using something like conda as far as I'm aware, so that can lead to confusion. So I agree we should remove them too.

PR #216, opened 2025-04-03
Add automatic downloads from OpenTopography 

### LINKED ISSUE #215: Add Automatic Downloads from OpenTopography

Teschl:
This adds two new functions to `utils.py`:
- `load_open_topography()`
- `read_from_cache()`

This adds the ability to Load DEM's from OpenTopography by specifying the bounds, global raster dataset and an API-Key. These will be cached to avoid having to download them again if called with the same arguments. There is also the option to save them at a desired path. To make it easier to load already cached datasets, the function `read_from_cache()` can get them from the cache. This way the user can avoid having to input the whole path, if they haven't saved the TIF at a different place using the `save_path` argument.

closes #215

### DIFF
--- a/src/topotoolbox/utils.py
+++ b/src/topotoolbox/utils.py
@@ -344,15 +344,18 @@ def read_from_cache(filename: str) -> GridObject:
 
 
 def load_open_topography(south: float, north: float, west: float, east: float,
-                         dem_type: str = "SRTMGL3", overwrite: bool = False,
-                         save_path: str | None = None,
-                         api_key: str = "demoapikeyot2021") -> GridObject:
+                         api_key: str | None = None, dem_type: str = "SRTMGL3",
+                         api_path: str | None = None, overwrite: bool = False,
+                         save_path: str | None = None
+                         ) -> GridObject:
     """Download a DEM from Open Topography. The DEM is downloaded as a
     GeoTIFF file and saved in the cache directory. The DEM is then
-    read into a GridObject. To overwrite an existing Downloaded DEM, use
-    the overwrite parameter. To save the DEM to a different location,
-    use the save_path parameter. The API key is required for accessing
-    the Open Topography API. The default API key is the demo key.
+    read into a GridObject. The DEMs come in geographic coordinates (WGS84).
+    The API key is required for accessing the OpenTopography API
+    (opentopography.org). To avoid having to pass the API key as an argument,
+    save it in .opentopography.txt in the working or home directory. To
+    overwrite an existing downloaded DEM, use the overwrite parameter. To save
+    the DEM to a different location, use the save_path parameter.
 
     Parameters
     ----------
@@ -364,6 +367,15 @@ def load_open_topography(south: float, north: float, west: float, east: float,
         WGS 84 bounding box west coordinates
     east : float
         WGS 84 bounding box east coordinates
+    api_key : str | None
+        The API key is needed to access the Open Topography API. To get an API
+        key, visit opentopography.org, log in, navigate to MyOpenTopoDashboard
+        and request a new API key. This argument is used to pass the
+        API key directly as a string. By default None.
+    api_path : str | None
+        The path to a file that contains the API key. If provided, the API key
+        will be read from the file. The file should contain only the API key
+        without any additional text or formatting. By default None.
     dem_type : str, optional
         Choose one of the available global raster types, by default "SRTMGL3"
 
@@ -385,12 +397,9 @@ def load_open_topography(south: float, north: float, west: float, east: float,
         If True cached DEM will be overwritten if it has the same bounds
         and dem_type, by default False
     save_path : str | None, optional
-        If provided, the downloadad GeoTIFF will be saved to this path. Like
-        this `"path/to/file.tif"` for example. By default None
-    api_key : str, optional
-        The API key for Open Topography. This is required to access the API.
-        If not provided, the demo key with limeted functionality will be used, 
-        by default "demoapikeyot2021"
+        If provided, the downloaded GeoTIFF will be saved to this path. Like
+        this `"path/to/file.tif"` for example. If it is None, files will
+        be saved in cache. By default None
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


Teschl:
We could also add the option of only deleting OpenTopography data from the cache to the `clear_cache` function. What do you think @wkearn ?

src/topotoolbox/utils.py
```diff
@@ -315,6 +322,141 @@ def get_cache_contents() -> (list[str] | None):
...
+    """
+    cache_path = os.path.join(get_save_location(), f"{filename}")
+    grid_object = read_tif(cache_path)
+    return grid_object
+
+
+def load_open_topography(south: float, north: float, west: float, east: float,
+                         dem_type: str = "SRTMGL3", overwrite: bool = False,
+                         save_path: str | None = None,
+                         api_key: str = "demoapikeyot2021") -> GridObject:
```
wkearn, line 349: Does this demo API key work? "demoapikeyot2022" works for me, but the 2021 one fails with an "Unauthorized" error.

wkearn, line 349: It might be better not to provide a default API key and to force users to obtain their own. We can fail with an error message that describes how to obtain one.

src/topotoolbox/utils.py
```diff
@@ -315,6 +322,141 @@ def get_cache_contents() -> (list[str] | None):
...
+
+    if save_path:
+        # Copy cached file to specified save path
+        with open(cache_path, 'rb') as src, open(save_path, 'wb') as target:
+            while True:
+                # 1 MB chunks
+                chunk = src.read(1_048_576)
+                if not chunk:
+                    break
+                target.write(chunk)
```
wkearn, line 454: Why do we save it to `cache_path` and then copy it to `save_path`? Is it possible to use the `save_path` if it is supplied and construct the cache path otherwise? This saves two copies of the DEM on the user's computer, which I think is not what users will expect if they supply a save path.

src/topotoolbox/utils.py
wkearn, line 349: Would it be possible to permit reading the API key from a file? The MATLAB version uses "opentopography.apikey" in the IOtools folder as a default, though that might need to change (TopoToolbox/topotoolbox3#9). [bmi-topography](https://pypi.org/project/bmi-topography/) uses `.opentopography.txt` either in the current directory or the user's home directory as a default, which seems reasonable. 

Both topotoolbox3 and bmi-topography use either a provided API key string or a fixed file name, but I feel like it would be nice to allow a user to specify either a key as a string or a filename, and to fall back on the default file name only if neither are provided. If that logic is too complicated, we could use the approach of the other packages.

src/topotoolbox/utils.py
```diff
@@ -315,6 +322,141 @@ def get_cache_contents() -> (list[str] | None):
...
+
+    overwrite : bool, optional
+        If True cached DEM will be overwritten if it has the same bounds
+        and dem_type, by default False
+    save_path : str | None, optional
+        If provided, the downloadad GeoTIFF will be saved to this path. Like
+        this `"path/to/file.tif"` for example. By default None
+    api_key : str, optional
+        The API key for Open Topography. This is required to access the API.
+        If not provided, the demo key with limeted functionality will be used,
```
wkearn, line 392: "limeted" -> "limited"

src/topotoolbox/utils.py
```diff
@@ -315,6 +322,141 @@ def get_cache_contents() -> (list[str] | None):
...
+        - EU_DTM (DTM 30m)
+        - GEDI_L3 (DTM 1000m)
+        - GEBCOIceTopo (Global Bathymetry 500m)
+        - GEBCOSubIceTopo (Global Bathymetry 500m)
+
+    overwrite : bool, optional
+        If True cached DEM will be overwritten if it has the same bounds
+        and dem_type, by default False
+    save_path : str | None, optional
+        If provided, the downloadad GeoTIFF will be saved to this path. Like
```
wkearn, line 388: "downloadad" -> "downloaded"

src/topotoolbox/utils.py
```diff
@@ -315,6 +322,141 @@ def get_cache_contents() -> (list[str] | None):
...
+def load_open_topography(south: float, north: float, west: float, east: float,
+                         dem_type: str = "SRTMGL3", overwrite: bool = False,
+                         save_path: str | None = None,
+                         api_key: str = "demoapikeyot2021") -> GridObject:
+    """Download a DEM from Open Topography. The DEM is downloaded as a
+    GeoTIFF file and saved in the cache directory. The DEM is then
+    read into a GridObject. To overwrite an existing Downloaded DEM, use
+    the overwrite parameter. To save the DEM to a different location,
+    use the save_path parameter. The API key is required for accessing
+    the Open Topography API. The default API key is the demo key.
```
wkearn, line 355: We should also say something about reprojecting the DEM into a projected coordinate system here. See the docs for [readopentopo](https://github.com/TopoToolbox/topotoolbox3/blob/main/toolbox/IOtools/readopentopo.m). Also see below for the example.

src/topotoolbox/utils.py
```diff
@@ -315,6 +322,141 @@ def get_cache_contents() -> (list[str] | None):
...
+        If the provided DEM type is not valid.
+    ConnectionError
+        If the API request fails or returns an error.
+
+    Example
+    -------
+    dem = topotoolbox.load_open_topography(south=50, north=50.1, west=14.35,
+                    east=14.6, dem_type="SRTMGL3", api_key="demoapikeyot2022")
+    im = dem.plot(cmap="terrain")
+    plt.show()
```
wkearn, line 412: To reproject this DEM, we need something like

```python
import rasterio

dem = topotoolbox.load_open_topography(south=50, north=50.1, west=14.35,
                    east=14.6, dem_type="SRTMGL3", api_key="demoapikeyot2022")
dem_utm = dem.reproject(rasterio.CRS.from_epsg(32633),resolution=90)
im = dem_utm.plot(cmap="terrain")
plt.show()
```

This won't work at the moment because `reproject` doesn't copy the bounding box, and the plot doesn't work without the bounding box. I'll fix that momentarily, though.

wkearn (changes requested):
This works well, @Teschl! Just a few small questions and fixes.

wkearn:
> We could also add the option of only deleting OpenTopography data from the cache to the `clear_cache` function. What do you think @wkearn ?

I think we can probably just rely on the `filename` argument to `clear_cache` for now. If someone really wants this in the future, it isn't too hard to add.

src/topotoolbox/utils.py
Teschl: I does work, I tested the function with the demo key. I assume it only supports a limited amount of bounds and/or dem_types. I used:

```python
load_open_topography(south=50, north=50.2, west=14.35, east=14.6, dem_type="SRTMGL3", api_key="demoapikeyot2022)"
 ```

Comment (Teschl) [reply src/topotoolbox/utils.py:349]: Thats probably a good idea :+1:

File: src/topotoolbox/utils.py (hunk 2 above)
Comment (Teschl) [reply src/topotoolbox/utils.py:454]: I will change it so it will only cache if not path is provided. I originally thought always caching would work well because it could eliminate downloading the same file twice. Like you said, that would be unexpected though.

File: src/topotoolbox/utils.py (hunk 1 above)
Comment (Teschl) [reply src/topotoolbox/utils.py:349]: Ah, I see. It's corrected now in the Docstring :+1:

File: src/topotoolbox/utils.py (hunk 2 above)
Comment (wkearn) [reply src/topotoolbox/utils.py:454]: That is a good point. You don't necessarily want to download the data multiple times. You might use `load_open_topography` in a script to make sure you have some data loaded. Perhaps throwing an error if `save_path` exists and `overwrite=False` is a good idea in that case -- make users explicitly set `overwrite=True` if they want to redownload the file.

File: src/topotoolbox/utils.py (hunk 1 above)
Comment (wkearn) [reply src/topotoolbox/utils.py:349]: > I think adding environment variables as another possibility would make sense too. Right now, I'm thinking the following options should work:
> 
>     * `api_key="someapikey"` passing the key directly
> 
>     * `api_key="file:path/to/somefile.txt"` choosing a file
> 
>     * `api_key="env:SOME_ENV"` choosing an environment variable
> 
> 
> Defaulting to `.opentopography.txt` would happen is it's just `api_key="file:"` and the default env `api_key="env:"` would be `OPENTOPOGRAPHY`

I don't think I like the prefixed string approach so much. It feels like it requires the same logic as passing separate keyword arguments, but now also requires us to parse a string. What about something like

```python
def load_open_topography(...,api_key=None, key_path=None):
    if api_key:
        key = api_key
    elif key_path:
        with open(key_path) as f:
            key = f.read().strip()
    else:
        key = os.getenv("OPENTOPOGRAPHY_API_KEY")
    
    if key is None:
            with open("~/.opentopography.txt") as f:
                key = f.read().strip()
    # ...
```

1. if the user supplies an API key directly, use it. 
2. If they don't supply a key but do supply a key path, use the key in that file. 
3. If they supply neither a key nor a key path, try to get an environment variable. This sets key to None if the environment variable is not set.
4. If none of the above succeeded, try to open a default file and get the key from it. If this file doesn't exist, the `open` will throw an error. We should probably catch that error and raise our own with an informative error message about setting the key properly.

> I think adding support for `.env` files would make sense: (this would mean another dependency, though)
> 
> ```python
> from dotenv import load_dotenv
> load_dotenv() 
> api_key = os.getenv("OPENTOPOGRAPHY")
> ```

I think this is something a user could choose to do to manage their environment variables. If we handle passing the key as an environment variable, this pattern will work for the user, but we don't need to explicitly depend on dotenv.

src/topotoolbox/utils.py
Teschl: This should be in the function, right? Not just the example?

wkearn: I was thinking just to have it in the example to remind users that they should probably reproject the DEM that you download. 

You need to know things like which CRS to reproject to, which would just add more complexity to the `load_open_topography` interface. The MATLAB `readopentopo` doesn't automatically reproject either.

Teschl:
I addressed the requested changes from your review @wkearn.
I decided against supporting environment variables because as it turns out they cause issues when being read from a Jupyter notebook. Getting this to work is an issue the user would have to solve, that we can't control from the package. We also can't catch if a missing environment variable is a user error, an issue in the package, or just the notebook not being able to find the variable. So I feel like it's simpler to just require the key as a .txt or string.

src/topotoolbox/utils.py
```diff
@@ -337,6 +531,6 @@ def validate_alignment(s1, s2) -> bool:
        True if the two objects are aligned, False otherwise
     """
     return (s1.shape == s2.shape) and all(
-        (not hasattr(s1,attr) or not hasattr(s2,attr))
-            or (getattr(s1,attr) == getattr(s2,attr))
-            for attr in ["bounds", "crs", "transform"])
+        (not hasattr(s1, attr) or not hasattr(s2, attr))
+        or (getattr(s1, attr) == getattr(s2, attr))
+        for attr in ["bounds", "crs", "transform"])
```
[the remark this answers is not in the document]
Teschl: Yes, I'm using the autopep8 formatter in VS Code. That should be `autopep8=2.3.2` when just installing it using pip.
I found the formatting of autopep8 to be better/less obstructing than the formatting of black or flake8. If you have a preference for one of these, it would also not be a big deal for me to switch to one of those.

wkearn:
Looks good!

Issue #187, opened 2025-03-20
Plot with coordinates


wschwanghart:
Currently, most plots seem to have number of rows and columns on the y and x-axis, respectively. Also, when I plot the planform layout of a STREAMobj, it will plot the rows and columns. Can we change the default behavior so that it plot the actual coordinates?

wkearn:
matplotlib uses [the  `origin` and `extent` arguments to `imshow`](https://matplotlib.org/stable/users/explain/artists/imshow_extent.html#imshow-extent) to set the axes. The `origin` says which corner the array indices start counting from and `extent` gives the bounding box in the desired coordinate system. This would be simpler with a utility function (akin to MATLAB's `wf2XY`) that computes the bounding box from the geotransform.

`FlowObject`s and `StreamObject`s both carry around the geotransform of their parent `GridObject`, so the same function could be applied for plotting other data structures.

PR #234, opened 2025-04-19
Add a first test for graphflood

bgailleton:
This PR adds a test case for graphflood, based on an analytical solution on a rectangular channel.

Note that:
- it's a "long" test (~5s), I will try to minimise it but it does require iterations and a grid not too small to avoid edge effects
- To work, it relies on the pull request [187](https://github.com/TopoToolbox/libtopotoolbox/pull/187) of `libtopotoolbox`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,8 +10,9 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 include(FetchContent)
 FetchContent_Declare(
   topotoolbox
-  GIT_REPOSITORY https://github.com/TopoToolbox/libtopotoolbox.git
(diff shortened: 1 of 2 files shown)


wkearn:
Right now, the test passes on CI with the older libtopotoolbox prior to merging TopoToolbox/libtopotoolbox#187. Should the test have failed without that change to libtopotoolbox? Does anything need to be changed here now that TopoToolbox/libtopotoolbox#187 is available?

bgailleton:
All good! It's ready to merge.
I forgot I restricted the `assert` on the top half of the DEM, unaffected by the boundary issue.
Basically, if there is anything seriously wrong (e.g. wrong inverted row/col indices) the final result would never be within the error margin anyway

PR #235, opened 2025-04-23
Support Python v3.12 and v3.13

wkearn:
This adds 3.12 and 3.13 to the CI test runners and 3.13 to the wheels that we build and submit to PyPI.

https://scientific-python.org/specs/spec-0000/ is a good resource for figuring out which versions of things we should support at a minimum.


wkearn:
Test failures are because Numpy only supports 3.12 from v1.26 and 3.13 from v2.1.0. We need to filter our Downgrade numpy step on the Python version.

PR #259, opened 2025-05-14
adding the to_shapefile feature

faurefrancois:
Here is the pull request of the to_shapefile feature, as mentionned in the corresponding issue report. Tell me if there is any problem!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,9 @@ build-backend = "scikit_build_core.build"
 requires = [
     "scikit-build-core",
     "pybind11",
-    "numpy>=2.0.0"
+    "numpy>=2.0.0",
+    "geopandas",
+    "shapely"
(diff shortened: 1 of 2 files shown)


wkearn:
I think you may have based your changes to `stream_object.py` on an earlier version of the `main` branch. It looks like there are a lot of changes that aren't related to your `to_shapefile` function that remove things we recently added.

The simplest way to fix this is probably with `git reset --mixed main`, which will reset your branch to `main`, but leave your changes in your files. Then change `stream_object.py` so that you have only added the `to_shapefile` function, stage and commit `stream_object.py`, and also recommit `pyproject.toml`. Then push again -- you'll need to do a force push to overwrite the older commits, but the new commits will show up here.

wkearn:
I will close this in favor of #260

Issue #258, opened 2025-05-14
Adding a to_shapefile function in the StreamObject class (QGIS useful)


faurefrancois:
![Image](https://github.com/user-attachments/assets/34028674-b7d7-4b30-9a11-35a941ef40d2)

Hello!

As I want to use the steam network in QGIS, I created a method that extracts the lines from the self.xy method and that generates a shapefile. I can then just import this shapefile in QGIS and it is directly overlapping the DEM (as, thanks to the crs attribute, we can georeference the shapefile).

Here is the method in the stream_object.py file, in the StreamObject class:

```
def to_shapefile(self, saving_path, saving_name):
        ''' Converts the stream network to a georeferenced shapefile
        
        Parameters
        ----------
        saving_path: python string
             path were the shapefile will be saved
            
        saving_name: python string
            name of the shapefile
        
        Returns
        ----------
        shapefile
            The georeferenced stream network shapefile
        
        '''
        line_geoms = [LineString(coords) for coords in self.xy()]
        gdf = gpd.GeoDataFrame(geometry=line_geoms, crs=self.crs)
        gdf.to_file(os.path.join(saving_path, saving_name))
```

with those libraries needed:
```
import os
from shapely.geometry import LineString
import geopandas as gpd

```

It can then be used like this:

```
import topotoolbox as tt3
import matplotlib.pyplot as plt

dem = tt3.read_tif('test.tif')
fd = tt3.FlowObject(dem)

s = tt3.StreamObject(fd,threshold=10000,units='pixels')
s.to_shapefile('/path/to/your/folder/', 'result')
```

It is the first time I try to contribute to a library so I am new to all this, please tell me if some stuff are not "best practices" and excuse me if my post is useless for the community (even if I hope not^^)

Have a good day,
François.

wkearn:
@faurefrancois: This looks great! We have similar functionality in MATLAB, so it is a great addition to the Python package!

Would you like to submit this as a pull request? We have some [guidance](https://topotoolbox.github.io/contributing.html#contributing-via-pull-request) for working through this process, and I can answer any additional questions that you have. Alternatively, I can add your patch myself.

faurefrancois:
@wkearn Thank you for your positive answer! I can try to add it myself, it would be a good thing to learn!

wkearn:
Resolved by #260

PR #260, opened 2025-05-15
Adding to_shapefile feature, issue with earlier version fixed (?)

faurefrancois:
I simply deleted the repo and clone a new one to modify just the needed lines. Here is a new push and it should be better.

One question:
I didn't modify the requirements.txt because I am not sure which version of shapely and geopandas are correct. How can I know this?

Also, shoud I add a short doc?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,8 +4,6 @@ requires = [
     "scikit-build-core",
     "pybind11",
     "numpy>=2.0.0",
-    "geopandas",
-    "shapely"
 ]
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,9 @@ build-backend = "scikit_build_core.build"
 requires = [
     "scikit-build-core",
     "pybind11",
-    "numpy>=2.0.0"
+    "numpy>=2.0.0",
+    "geopandas",
+    "shapely"
 ]
 
 
@@ -23,7 +25,9 @@ dependencies = [
     "numpy>=1.23.5",
     "matplotlib",
     "scipy",
-    "rasterio"
+    "rasterio",
+    "geopandas",
+    "shapely"
 ]
 requires-python = ">=3.10"
 keywords = ["TopoToolbox"]
--- a/src/topotoolbox/stream_object.py
+++ b/src/topotoolbox/stream_object.py
@@ -1,5 +1,6 @@
 """This module contains the StreamObject class.
 """
+import os
 import math
 import warnings
 import copy
@@ -8,6 +9,8 @@ import numpy as np
 import matplotlib.pyplot as plt
 from matplotlib.collections import LineCollection
 from scipy.sparse import csr_matrix
+from shapely.geometry import LineString
+import geopandas as gpd
 
 from .flow_object import FlowObject
 from .grid_object import GridObject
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -3,7 +3,9 @@ build-backend = "scikit_build_core.build"
 requires = [
     "scikit-build-core",
     "pybind11",
-    "numpy>=2.0.0"
+    "numpy>=2.0.0",
+    "geopandas",
+    "shapely"
```
wkearn, line 8: This section is for packages necessary when we compile our C/C++ code. You probably don't need to add geopandas and shapely here.

wkearn:
> I simply deleted the repo and clone a new one to modify just the needed lines. Here is a new push and it should be better.
:+1: 

> One question: I didn't modify the requirements.txt because I am not sure which version of shapely and geopandas are correct. How can I know this?

This is a good question. I think at the moment we just need a version that we know works with the other packages in `requirements.txt`, so just listing `geopandas` and `shapely` without versions should be all right. If we start getting errors, we can specify things more explicitly.

We use `requirements.txt` only for setting up developer environments. We ran into problems at one point where different developers were using slightly different versions of sphinx, and the documentation came out looking wrong.

> Also, shoud I add a short doc?

Do you mean a separate tutorial that goes in the documentation site? You can if you would like to. You do already have the docstring for the function, which will show up in the API reference. You can add an "Example" section to your docstring -- see some of the other functions for formatting -- and put something like the example that you gave in #258:

```python
dem = tt3.load_dem('bigtujunga')
fd = tt3.FlowObject(dem)
s = tt3.StreamObject(fd)
s.to_shapefile('.', 'stream_network.shp')
```

just using the built-in example DEM so that it will work for other users.

src/topotoolbox/stream_object.py
```diff
@@ -385,6 +388,26 @@ def xy(self, data=None):
                         stack.append(v)
 
         return segments
+    
+    def to_shapefile(self, saving_path, saving_name):
```
wkearn, line 392: You can combine the `saving_path` and `saving_name` arguments into a single string argument. See, for example, the `write_tif` function: https://github.com/TopoToolbox/pytopotoolbox/blob/5dbc1400f37dc64c6f89b60504cfe4607c962fe2/src/topotoolbox/utils.py#L29

Then you will not need the `os` import either.

wkearn, line 392: Would it be better to provide a method that converts a `StreamObject` to a `GeoDataFrame` and then let the user call `to_file` on the returned object when they want to save it? I can imagine converting the `StreamObject` to a `GeoDataFrame` could be useful on its own if a user wanted to analyze it further within Python, or convert it to another object like a GeoJSON string.

This is also how MATLAB TopoToolbox does it, converting the `STREAMobj` into MATLAB's geotable, which you can then choose to write to a shapefile.

wkearn (changes requested):
@faurefrancois: This looks good! There are some minor details with `pyproject.toml` and `requirements.txt` that should be addressed.

There is a bigger picture question about whether we should provide an interface that writes to a shapefile or just returns the GeoDataFrame. It is good to figure this out now before we commit to providing the `to_shapefile` interface, which might be hard to remove or change in the future once people start using it.

src/topotoolbox/stream_object.py
wschwanghart: We could have both. A more versatile function to convert a `StreamObject` to a `GeoDataFrame`, and a function to easily and rapidly export a `StreamObject` to a shapefile without options regarding attributes and feature lengths.

wkearn (approved):
Excellent! Thanks for the contribution, @faurefrancois!

PR #278, opened 2025-05-31
Add resize and zscore

### LINKED ISSUE #264: Interactive GridObject resize function
A function that can be used to "zoom" into an area of a DEM. Would return a new GridObject cut down to the selected area. This makes sense to implement in a way where the selection happens on a plot of the DEM (instead of passing the area through arguments). While I think this would make a lot of sense as a function, implementing it showcased a few difficulties.

- Backend: To be able to use interactive plots, we need an [interactive backend](https://matplotlib.org/stable/users/explain/figure/backends.html). 
    - `TkAgg` uses TkInter which should come pre-installed on windows and mac, but not on every linux OS. So, using this would require installing py-tkinter (`sudo apt install python3-tk` not with pip!).
    - `QtAgg` needs to be installed with pip (`pip install PyQt6`)
    
    Since we want to provide full functionality in jupyter notebooks, both of these are flawed.
    - `ipympl` provides support for notebooks, but makes for standalone scripts harder to implement

- Getting functions to work in .ipynb is very hard
  [In this notebook](https://github.com/Teschl/pytopotoolbox/blob/interactive/rectangle_selector.ipynb.ipynb) I implemented a bare-bones version to be able to select an area to be the new DEM. While getting the selector to work is easy, working with the selected area is very hard (in notebooks, in scripts it's very straight forward). When this is implemented in a function that returns something, it will return as soon as the plot is done, so before the user can select an area. The only way I found to get around that is to use global variables.

```python
selected_region = data[i1:i2, j1:j2]
global saved_region
saved_region = selected_region
```

As you can see here, the ndarray get resized but has to be turned into a global variable. Meaning if this was used in a function, it would look something like this:

```python
dem = tt.load_dem('perfectworld')
dem.resize()

# select area ...
# new notebook cell

new_dem.plot()
```

But new_dem is not a user defined name, it would be hard coded in the function. Another drawback is, that there is no feedback on when the new DEM has been created.

I'm not sure how to proceed from here. Making this work in notebook seems to be impractical, but it would also be bad if the resize function can only be used in scrips (.py files). A solution could be, to create an interactive function that can be used to get the new border coordinates (but only in scripts) and another function to actually resize, but only using coordinates. Its probably better to hard-code coordinates when resizing in notebooks anyway.

What do you think @wkearn ?

Teschl:
This adds the functions `zscore` and `resize` to the `GridObject`.

Maybe `crop` would be a more fitting name for the `resize` function. Also, having the arguments `left`, `right`, `top` and `bottom` might be unnecessary. The function only needs two x and y values each and figures out which is left/right or top/bottom automatically. Using `highlight_selected` will create a plot which highlights the area the new `GridObject` will be cropped to. 
I am not 100% sure the updating of the bounds and transform is completely correct, so if anything seems off about that please let me know.
Closes #264

### DIFF
--- a/src/topotoolbox/grid_object.py
+++ b/src/topotoolbox/grid_object.py
@@ -1252,6 +1252,12 @@ class GridObject():
         result.z = (self.z - np.nanmean(self.z)) / np.nanstd(self.z)
         return result
 
+    def castshadows(self, azimuth, altitude) -> 'GridObject':
+        pass
+
+    def impaintnans():
+        pass
+
         # 'Magic' functions:
         # ------------------------------------------------------------------------
 
--- a/src/topotoolbox/grid_object.py
+++ b/src/topotoolbox/grid_object.py
@@ -170,7 +170,8 @@ class GridObject():
             dst.cellsize = abs(dst.transform[0])
 
         if self.bounds:
-            dst.bounds = BoundingBox(*transform_bounds(self.crs, dst.crs, *self.bounds))
+            dst.bounds = BoundingBox(
+                *transform_bounds(self.crs, dst.crs, *self.bounds))
 
         return dst
 
@@ -270,7 +271,7 @@ class GridObject():
         if output is None:
             output = ['sills', 'flats']
 
-        dem = np.asarray(self,dtype=np.float32)
+        dem = np.asarray(self, dtype=np.float32)
         output_grid = np.zeros_like(dem, dtype=np.int32)
 
         _grid.identifyflats(output_grid, dem, self.dims)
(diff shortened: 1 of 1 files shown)


wkearn (changes requested):
`zscore` looks good! Happy to merge this now if you'd like to split it off into another PR, or we can wait until we figure out `resize`.

`resize` works great. There is some subtlety about the transform/bounds calculation that should be addressed, and I have a question about the plots.

Teschl:
I implemented the requested changes. I added the plotting a rectangle example to the docstring since I was able to compress it into a relatively small space.
Closes #264

wkearn (approved):

PR #298, opened 2025-06-11
fixed upstream_distance error

ginaarnau:


wkearn:
It turns out that we don't have a max-add traversal implemented in libtopotoolbox yet. We will need to do that before we can fix this error.

PR #308, opened 2025-07-01
Rename resize to crop

Teschl:
This implements the changes suggested by @wschwanghart [here](https://github.com/TopoToolbox/pytopotoolbox/commit/da15efdad8f092700294cba0eaadb76a20f864ab#commitcomment-160934428).

- Renamed resize to crop
- updated example in docstring
- Added `mode` argument to manually select crop method
- if no mode is provided, function will automatically select using the old logic (let me know if this should also be removed)

### DIFF
--- a/src/topotoolbox/grid_object.py
+++ b/src/topotoolbox/grid_object.py
@@ -1403,39 +1403,34 @@ class GridObject():
 
     def crop(self, left: float | int, right: float | int,
                top: float | int, bottom: float | int,
-               mode: str | None = None) -> 'GridObject':
+               mode: str) -> 'GridObject':
         """Crop the Gridobject by specifying new boundaries.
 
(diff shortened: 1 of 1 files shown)


wkearn (commented):
I think it is better to require users to supply the `mode` parameter. It presents some friction to the user, but trying to guess what a user means does seem likely to cause errors or confusion. 

I also don't have a good sense of what would be a good default here. I would probably use coordinates if I was trying to crop a DEM to a region of interest based on some other data, but pixels if I was just trying to restrict my analysis to a part of the DEM, and percentages for visualizations. In the absence of a good default, I would lean on the side of having no default (and requiring users to pass the `mode` parameter) rather than choosing one arbitrarily.

Teschl:
I agree, having no set default when removing the automatic detection makes the most sense. I implemented the changes, should be ready to merge now.

PR #318, opened 2025-07-11
changed coordinate system crs to georef

ginaarnau:


src/topotoolbox/grid_object.py
```diff
@@ -181,9 +181,9 @@ def reproject(self,
         z, dst.transform = reproject(
             self.z,
             src_transform=self.transform,
-            src_crs=self.crs,
+            src_georef=self.georef,
             dst_transform=None,  # Let rasterio derive the transform for us
-            dst_crs=crs,
+            dst_georef=georef,
```
wkearn, line 186: The argument names to this function are `src_crs` and `dst_crs`, so those need to be left as is:

```python
# ...
src_crs=self.georef,
# ...
dst_crs = georef,
# ...
```

ginaarnau: for some reason when changing that, all these  errors are appearing relating the stresm_object: 

src\topotoolbox\stream_object.py:88: error: Cannot determine type of "cellsize"  [has-type]
src\topotoolbox\stream_object.py:89: error: Cannot determine type of "shape"  [has-type]
src\topotoolbox\stream_object.py:90: error: Cannot determine type of "strides"  [has-type]
src\topotoolbox\stream_object.py:94: error: Cannot determine type of "bounds"  [has-type]

Multiple more errors like these (I just put the first few ones)

wkearn: This is a problem that sometimes occurs with one of the pre-commit hooks. It only fails in the pre-commit hook when you change something in grid_object.py. It won't fail on the regular GitHub Actions. As long as you have no other errors, you can bypass the hooks to commit your change (Commit (No Verify) in VS Code) and push to GitHub, and it should work.

src/topotoolbox/grid_object.py
```diff
@@ -141,19 +141,19 @@ def astype(self, dtype):
...
-                  crs: 'CRS',
+                  georef: 'CRS',
                   resolution: 'float | None' = None,
                   resampling: 'Resampling' = Resampling.bilinear):
         """Reproject GridObject to a new coordinate system.
 
         Parameters
         ----------
-        crs : rasterio.CRS
+        georef : rasterio.georef
```
wkearn, line 156: The name of the argument is `georef`, but the type is `rasterio.CRS`.

src/topotoolbox/grid_object.py
```diff
@@ -172,7 +172,7 @@ def reproject(self,
         -------
         >>> dem = topotoolbox.load_open_topography(south=50, north=50.1, west=14.35,
                     east=14.6, dem_type="SRTMGL3", api_key="demoapikeyot2022")
-        >>> dem = dem.reproject(rasterio.CRS.from_epsg(32633), resolution=90)
+        >>> dem = dem.reproject(rasterio.georef.from_epsg(32633), resolution=90)
```
wkearn, line 175: This should stay `CRS` because it is coming from rasterio.

src/topotoolbox/utils.py
```diff
@@ -103,7 +103,7 @@ def read_tif(path: str) -> GridObject:
         grid.cellsize = dataset.res[0]
         grid.bounds = dataset.bounds
         grid.transform = dataset.transform
-        grid.crs = dataset.crs
+        grid.georef = dataset.georef
```
wkearn, line 106: This should be

```python
grid.georef = dataset.crs
```

The `dataset` comes from rasterio, so it calls its coordinate information `crs`.

wkearn (changes requested):
There are a few places where you changed some instances of `crs` that belong to rasterio. Those need to be switched back to `crs`. I am pretty sure I noted all of these in line comments.

PR #324, opened 2025-07-22
added doctest in stream, grid, and flow objects, and utils

ginaarnau:


src/topotoolbox/utils.py
```diff
@@ -460,10 +478,12 @@ def load_opentopography(south: float, north: float, west: float, east: float,
...
-    >>> dem = topotoolbox.load_open_topography(south=50, north=50.1, west=14.35,
-                    east=14.6, dem_type="SRTMGL3", api_key="demoapikeyot2022")
-    >>> dem = dem.reproject(rasterio.CRS.from_epsg(32633), resolution=90)
-    >>> im = dem.plot(cmap="terrain")
+    >>> import topotoolbox
+    >>> import matplotlib.pyplot as plt
+    >>> from rasterio import CRS
+    >>> dem = topotoolbox.load_dem('kunashiri')
+    >>> dem = dem.reproject(CRS.from_epsg(32655), resolution=30)
+    >>> _= dem.plot(cmap="terrain")
```
wkearn, line 486: This example is supposed to demonstrate the `load_opentopography` function, so we can't replace it with loading the geographic DEM like in the `reproject` example. It is better to leave it as

```python
>>> import topotoolbox
>>> import matplotlib.pyplot as plt
>>> from rasterio import CRS
>>> dem = topotoolbox.load_dem('kunashiri')
>>> dem = dem.reproject(CRS.from_epsg(32655), resolution=30)
>>> dem = topotoolbox.load_opentopography(south=50, north=50.1, west=14.35,
...     east=14.6, dem_type="SRTMGL3")
>>> dem = dem.reproject(CRS.from_epsg(32633), resolution=90)
>>> _= dem.plot(cmap="terrain")
>>> plt.show()
```

Also note that I got rid of the demo API key here. We shouldn't use that for anything that we are going to run regularly. If we run the test on GitHub Actions, we can use the OpenTopography API key that we have stored in the repository.

If you run into problems with this test, either because you don't have an API key or because you have problems accessing the DEM over the network, you can skip this example by adding `# doctest: +SKIP` to each line that you don't want to run:

```python
>>> import topotoolbox
>>> import matplotlib.pyplot as plt
>>> from rasterio import CRS
>>> dem = topotoolbox.load_dem('kunashiri')
>>> dem = dem.reproject(CRS.from_epsg(32655), resolution=30)
>>> dem = topotoolbox.load_opentopography(south=50, north=50.1, west=14.35,
...     east=14.6, dem_type="SRTMGL3") # doctest: +SKIP
>>> dem = dem.reproject(CRS.from_epsg(32633), resolution=90) # doctest: +SKIP
>>> _= dem.plot(cmap="terrain") # doctest: +SKIP
>>> plt.show()
```

wkearn (commented):
Looks pretty good. There is one issue with the `load_opentopography` test detailed in a comment there.

Can you also add the `doctest` extension to `docs/conf.py`? This will let us use `doctest` in our documentation build process. Add a line to the `extensions` list at the top of the file:

```diff
@@ -9,6 +9,7 @@ import topotoolbox
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.autosummary',
+    'sphinx.ext.doctest',
     'sphinx.ext.napoleon',
     'sphinx.ext.viewcode',
     'sphinx.ext.todo',
```

src/topotoolbox/utils.py
wkearn: I think the issues with `crs` and `crslin` are mostly not fatal errors related to the docstring format.

The real error is the `filename or extension is too long` error. It looks like sphinx is running recursively, generating a directory inside a directory and then trying to rerun from there, so you eventually get a filename that is too long.

How are you running sphinx? I think when we added the `sphinx.ext.doctest` extension to `conf.py` it might have changed how you need to run things. With the extension, running

```bash
make doctest
```

from a terminal (not a Python interpreter but the bash or Powershell terminal where you can run `git` commands) should work.

wkearn: You probably don't have `make` installed. That's all right. It is actually just running the same command you are using, so I don't think that's the problem, necessarily.

wkearn:
Closed in favor of #403, which updates this PR with recent changes to pytopotoolbox.

PR #326, opened 2025-07-23
changed crslin to use clarabel instead of kvxopt

ginaarnau:


src/topotoolbox/stream_object.py
```diff
@@ -1215,46 +1219,48 @@ def crslin(self, dem, k = 1, mingradient=0.0, attachheads=False, attachtomin=Fal
...
-            nr, nr)) - sp.coo_matrix((dd, (self.source, self.source)), shape=(nr, nr))).tocoo()
-        gradient = spmatrix(gradient.data.tolist(), gradient.row.tolist(
-        ), gradient.col.tolist(), (nr, nr))  # convert to ckxopt matrix
+        gradient = sp.csc_array(
+            (delta, (np.arange(ne), self.target)),shape=(ne, nr)) - sp.csc_array(
+                (delta, (np.arange(ne), self.source)), shape=(ne, nr))
 
-        g_min = np.zeros((nr, 1))  # minimum gradient
+        g_min = np.zeros(nr)  # minimum gradient
         g_min[self.source] = mingradient
```
wkearn, line 1239: The remaining errors are because `g_min` needs to have length `ne`, but you have left it the older way with `nr` elements. Change this line to:

```python
g_min = np.full(ne, mingradient)
```

src/topotoolbox/stream_object.py
```diff
@@ -1186,26 +1186,30 @@ def crslin(self, dem, k = 1, mingradient=0.0, attachheads=False, attachtomin=Fal
         values = np.array(
             [2/((xi-xj)*(xk-xj)), -2/((xk-xi)*(xi-xj)), 2/((xk-xi)*(xk-xj))])
 
-        # Sparse kvxopt second derivative matrix (C)
-        c = spmatrix(values.T.flatten().tolist(), rowix.flatten(
-        ).tolist(), colix.flatten().tolist(), (nrrows, nr))
+        # Second derivative matrix (C)
+        c = sp.csc_array((values.flatten(), (rowix.flatten(), colix.flatten())), (nrrows, nr))
```
wkearn, line 1190: You dropped the needed transpose when you converted this to `csc_array`. You need the change:

```diff
@@ -1187,7 +1187,7 @@ class StreamObject():
             [2/((xi-xj)*(xk-xj)), -2/((xk-xi)*(xi-xj)), 2/((xk-xi)*(xk-xj))])
 
         # Second derivative matrix (C)
-        c = sp.csc_array((values.flatten(), (rowix.flatten(), colix.flatten())), (nrrows, nr))
+        c = sp.csc_array((values.T.flatten(), (rowix.flatten(), colix.flatten())), (nrrows, nr))
 
         #########################################
         # Compute s parameter (equation A7)
```

src/topotoolbox/stream_object.py
```diff
@@ -1186,26 +1186,30 @@ def crslin(self, dem, k = 1, mingradient=0.0, attachheads=False, attachtomin=Fal
...
+        b = np.concatenate([(z.reshape(-1, 1)), np.zeros((nrrows, 1))])
 
         # Compute function parameters for quadratic programming (equation A11)
-        f = -2*(a_matrix.T * b)
-        h_matrix = 2*(a_matrix.T * a_matrix)
+        f = -2*(a_matrix.T @ b)
+        f = f.flatten()
+        h_matrix = 2*(a_matrix.T @ a_matrix).tocsc()
 
+        #########################################
```
wkearn, line 1212: Here `b` is a two-dimensional matrix with a single column, and then you flatten `f` to turn it into a vector. It is a bit more concise to create `b` as a vector:


```diff
@@ -1202,11 +1202,10 @@ class StreamObject():
         # Compute matrix A and vector b (equation A9)
         a_2 = s_parameter*c  # C*s
         a_matrix = sp.vstack([identity_matrix, a_2], format = 'csc')
-        b = np.concatenate([(z.reshape(-1, 1)), np.zeros((nrrows, 1))])
+        b = np.concatenate([z, np.zeros(nrrows)])
 
         # Compute function parameters for quadratic programming (equation A11)
         f = -2*(a_matrix.T @ b)
-        f = f.flatten()
         h_matrix = 2*(a_matrix.T @ a_matrix).tocsc()
 
         #########################################
```

src/topotoolbox/stream_object.py
```diff
@@ -1215,46 +1219,47 @@ def crslin(self, dem, k = 1, mingradient=0.0, attachheads=False, attachtomin=Fal
...
+        h = np.concatenate([z_eq, h])
 
         # Solve quadratic programming with the constraints
-        zs = qp(h_matrix, f, m_matrix, h, i_eq, z_eq,
-                options={'show_progress': False})
-        zs = np.array(zs['x']).flatten()
+        cones = [ZeroConeT(i_eq.shape[0]), NonnegativeConeT(m_matrix.shape[0])]
+
+        settings = DefaultSettings()
+        settings.verbose = True
```
wkearn, line 1257: This should be False so we don't get the messages from clarabel printed.

wkearn (changes requested):
To get the tests to pass, I think you will also need to change

```diff
@@ -628,7 +628,7 @@ def test_crslin(wide_dem, wide_stream):
 
     # attachheads: zs(channelheads) = z(channelheads)
     channelheads = s.streampoi('channelheads')
-    assert np.all(zs[channelheads] == z[channelheads])
+    assert np.allclose(zs[channelheads], z[channelheads])
 
     # gradient: gradient > mingradient
     gradient = (zs[s.source] - zs[s.target]) / s.distance()
```
because the constraint doesn't exactly hold in the clarabel solution.

wkearn (approved):
Great work, @ginaarnau!

PR #343, opened 2025-08-14
Add GraphFlood example

wkearn:
This adds a Jupyter notebook that runs GraphFlood on the Green River DEM as in Gailleton et al. (2024). This generates the figure that is planned for the TT3 manuscript.

@bgailleton: Any ideas for improving this example?

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -4,3 +4,4 @@ Examples
 .. nbgallery::
    /examples/magicfunctions
    /examples/excesstopography
+   /examples/GraphFlood
--- a/docs/examples/GraphFlood.ipynb
+++ b/docs/examples/GraphFlood.ipynb
@@ -1,0 +1,98 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "0",
+   "metadata": {},
+   "source": [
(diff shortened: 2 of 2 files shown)


bgailleton:
@wkearn I have two main suggestions. First it can be interesting to have a bit more outputs than just the flow depth, for example flow velocity or discharge or eventually other component that would demonstrate the advantage of using both graphflood integrated into topotoolbox. 
Q and u are not directly calculable with the `run_graphflood` method. I am thinking of completing the interface of graphflood on pytopotoolbox with a Object Oriented interface to facilitate the use, for example 
``` python
gfo = GraphFloodObject(...)
gfo.run(N_iterations)
gfo.get_Qin()
gfo.run(N_more_iterations)
gfo.get_u()
...
```

Second suggestion, I do not mind using green_river, but as it has already been used in the 2024 paper, maybe another field site would bring more demo?

let me know your thoughts.

bgailleton:
#344 added the graphflood Object and paves the way to the example (sorry for the delay)

wkearn:
A more complete GraphFlood example has been added to the gallery (TopoToolbox/gallery#31), so this is somewhat obsolete.

PR #344, opened 2025-09-21
Graphflood obj update

bgailleton:
# Add Graphflood Object (GFObject) + update

## GFObject

This PR adds a brand new Graphflood Object to manage GraphFlood simulation.

Helps with running the model from scratch, checking convergence, run for longer if needed, ...

It also provides some analysis tools: extract DIscharge in and out (drainage area based vs computed - will be equal if model has converged but can be used for hillslopes that converge very slowly but does not move much in term of flow partitions), flow velocity, discharge per unit width or even river shear stress.

## Update

Recent graphflood's PR on libtopotoolbox brought a couple of changes into the API and this PR updates pytopotoolbox to it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ include(FetchContent)
 FetchContent_Declare(
   topotoolbox
   GIT_REPOSITORY https://github.com/TopoToolbox/libtopotoolbox.git
-  GIT_TAG 1084fc42c4e8428829d56e5f3d3c18cea3d6a906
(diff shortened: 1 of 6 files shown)


bgailleton:
Ok I think it still points to libtopotoolbox  without the graphflood update.
I just need to change the tag in the CMakeList or will it be automatic?

wkearn:
You will need to change the libtopotoolbox tag in the CMakeLists.txt to one 
that includes the graphflood updates.

PR #355, opened 2025-12-08
View after crop

bgailleton:
After cropping a `GridObject` `(GridObject.crop(...)`), the returned GridObject is a view into the parent, which then fails when passing to a `libtopotoolbox` function (e.g. fillsinks).

Here I simply added an explicit copy rather than a view, @wkearn is it OK with you?


wkearn:
Making an explicit copy is fine. I think that is the behavior we want here.

A longer term goal is to make libtopotoolbox handle views as well. There is 
a long-standing issue on libtopotoolbox (TopoToolbox/libtopotoolbox#23) for 
supporting strided arrays, which would make this possible.

It looks like `GridObject.resample` got removed in your commit, though. Was 
that intentional?

PR #361, opened 2026-01-30
Relative Elevation above rivers

bgailleton:
Add two  features:
- in `StreamObject`: a property `gridmask` that returns a grid-shape mask of where river nodes are
- in `FlowObject`: a function `relative_elevation_to_stream` that returns the relief between a node and the closest river node in flow path

Note: The flow object has a local import within the function to check the type of StreamObject without triggering circular import, and I had to add a pylint exception.


wkearn:
Yes, just so we have all the tests green before merging.

Issue #210, opened 2025-04-03
Implement `FlowObject.ezgetnal`


wkearn:
Discussed in #194. This provides an interface for accessing values of arrays through a `FlowObject` that is symmetric to that of `StreamObject`. It should check alignment of the supplied array or `GridObject` (#209) and return its argument, either an array or a `GridObject` as needed. This should be a method on `FlowObject`.

PR #365, opened 2026-02-20
Binding swath tools

bgailleton:
This PR bonds the swath tools recently PRed to `libtopotoolbox`:

**https://github.com/TopoToolbox/libtopotoolbox/pull/210**

### DIFF
--- a/src/topotoolbox/swath.py
+++ b/src/topotoolbox/swath.py
@@ -340,8 +340,9 @@ def _pca_tangent(ti, tj, pt, half_n):
     cij = float(np.dot(di, dj))
     cjj = float(np.dot(dj, dj))
     diff = cii - cjj
-    D = np.sqrt(diff * diff + 4.0 * cij * cij)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -21,6 +20,8 @@ find_package(pybind11 CONFIG REQUIRED)
 pybind11_add_module(_grid src/lib/grid.cpp)
 target_link_libraries(_grid PRIVATE topotoolbox)
 
+pybind11_add_module(_flow src/lib/flow.cpp)
+target_link_libraries(_flow PRIVATE topotoolbox)
```
wkearn: Not sure if you noticed, but I removed `src/lib/flow.cpp` in #371, so it needs to be removed from here and from the `install` step below.

bgailleton: Yes, I was still using my local CMakeList, should be fiixed now!

PR #439, opened 2026-05-26
Add a FlowObject.grid method

wkearn:
This method is meant to be similar to the GRIDobj(FD, Z) constructor in MATLAB. It is added as a FlowObject method, though, because the source flow object is where all of the metadata comes from.

It copies metadata from the source flow object and then stores the provided array.

Any keyword arguments are forwarded to numpy.asarray. This is simpler than including all the possible keyword arguments to asarray, which would also require additional maintenance if such arguments change.

A test is added that runs FlowObject.grid with a few different keyword arguments and makes sure that copies are created or not and that the data types are as expected.

The FlowObject methods that produce
GridObjects (e.g. flow_accumulation, drainagebasins) now use FlowObject.grid to construct their result.

A side effect of this change is that FlowObject.influencemap and FlowObject.dependencemap can now accept numpy arrays rather than GridObjects. Previously, they used the source GridObject to construct their results. This would error if a raw numpy array was passed. It is quite common to construct such an array as a mask, and it is a little annoying to remember to call `duplicate_with_new_data` before calling `influencemap`. The {influencemap, dependencemap}_indexing tests are updated to use a raw numpy array as an input to make sure that this functionality works as expected.

### DIFF
--- a/src/topotoolbox/flow_object.py
+++ b/src/topotoolbox/flow_object.py
@@ -329,6 +329,43 @@ class FlowObject():
 
         return k.astype(dtype)
 
+    def grid(self, z, **kwargs):
+        """Construct a GridObject aligned with the FlowObject from an array of data.
+
+        Keyword arguments are forwarded to a call to
+        `numpy.asarray`. Use these keyword arguments to specify the
+        data type (`dtype`) or whether to make a copy (`copy: bool`),
+        for example.
+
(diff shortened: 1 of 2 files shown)


wkearn:
Looks like GitHub Actions is having some trouble today: https://www.githubstatus.com/?today

PR #446, opened 2026-05-28
PPS: Implement rhohat

wkearn:
This PR provides an implementation of the rhohat estimator. Not all of the functionality of the MATLAB version is supported. Plotting is, as usual, left to the user, and we return just the estimates of the intensity and its confidence intervals. The ecdf function that is used in rhohat is replaced by a ksdensity function that only does the kernel density estimation required by rhohat. The ecdf function can be added separately if necessary (the MATLAB ecdf function may try to do too many things). ksdensity is also simplified by having it return all of the bootstrap samples. Weighted datasets are not yet supported.

At the moment, the bandwidth selected by scipy.stats.gaussian_kde by default is somewhat larger than that selected by MATLAB, especially for small datasets.

### DIFF
--- a/src/topotoolbox/pps.py
+++ b/src/topotoolbox/pps.py
@@ -7,6 +7,8 @@ import matplotlib.pyplot as plt
 import numpy as np
 import numpy.typing as npt
 
+import scipy.stats as st
+
 from .stream_object import StreamObject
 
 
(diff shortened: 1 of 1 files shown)


wschwanghart:
Very nice, @wkearn. If you have suggestions for changing and improving the MATLAB version, this is very welcome. The current PPS tools are not at the core of TT, so there would a lower probability to break things if we make some substantial changes.

PR #449, opened 2026-07-01
Use actions from the TopoToolbox/actions repository

wkearn:
We don't use a version number/commit SHA because we control that from
the TopoToolbox/actions repository.

The other workflows have not been updated to make sure that this
works.


wkearn:
Submodules may not be the right way to go with this. I think GitHub Actions pulls the repository without instantiating the submodules and so it gets empty directories for each of the actions.

